PR #39, opened 2023-01-12
feat: background service worker as RPC server

### LINKED ISSUE #6: Message bridge for Nexus
**Is your feature request related to a problem? Please describe.**

The injected CKB Provider will communicate with the Nexus, so a message bridge (wrapper) is required for Nexus. This bridge should 

- webpage <-> content script
- content script <-> background script
- background script <-> extension UI, e.g. `chrome.window` created window, popup.

For the CKB Provider, it may act as an RPC client, so we also need to make the bridge simply adaptable to server-client mode

```js
// perhaps the CKB provider will be like this
await window.ckb.nexus.getUnusedLocks()
await window.ckb.nexus.getCells()

// or act as RPC client
await window.ckb.nexus.request({
  method: 'wallet_getUnusedLocks',
  params: []
})
await window.ckb.nexus.request({
  method: 'wallet_getCells',
  params: []
})
```

**Additional context**

[webext-bridge](https://github.com/zikaari/webext-bridge) looks good, but webext-bridge seems not supported for injected script, maybe we can enhance it

[chomex](https://github.com/otiai10/chomex) offers server-client functionality, we can also have a look to check if we can implement it based on

homura:
Resolve #6, to try this PR, we can follow the [README](https://github.com/ckb-js/nexus/blob/554df786d297e7cf22f55353f90f7e5015a8b341/packages/extension-chrome/README.md)

```sh
git clone --branch json-rpc-2.0 --recurse-submodules https://github.com/ckb-js/nexus.git
cd nexus
npm install
npm run build
npm run test

cd packages/extension-chrome
npm start
```

This PR is a scaffold that treat the background script as an RPC server, and we can add the corresponding method by adding it to `addMethod`, like this

```ts
addMethod('wallet_enable', async (_, context) => {
  const { url, favIconUrl } = await context.getRequesterAppInfo();
  const { accessService, notificationService } = context.services;

  const accessible = await accessService.getAccessible({ url });
  if (accessible) return;

  const approved = await notificationServices.requestUserToApproveEnableWallet({ url, favIconUrl });
  if (approved) {
    await context.accessService.grantAccess({ url });
    return;
  }

  throw new Error('User rejected');
});

addMethod('wallet_fullOwnership_getUnusedLocks', () => {
  // mock
  return [{
      codeHash: '0x0000000000000000000000000000000000000000000000000000000000000000',
      hashType: 'type',
      args: '0x0000000000000000000000000000000000000000',
  }]
});
```

To ensure type safe, we need to define the interface before we implement a method

```ts
export interface RpcMethods {
  wallet_enable: RpcCall<void, void>;
  wallet_isEnabled: RpcCall<void, boolean>;
  wallet_getNetworkName: RpcCall<void, string>;

  wallet_fullOwnership_getUnusedLocks: RpcCall<void, Script[]>;
}

interface ServerParam {
  getRequesterAppInfo(): Promise<{ url: string; favIconUrl: string }>;
}
```

And we can try to call RPC methods like this in a console

```ts
const provider = await ckb.enable();
// unimplemented
await provider.fullOwnership.getUnusedLocks()
```

image

---

This PR will probably need to be refactored. There doesn't seem to be an existing messaging solution that fit our requirement too well, since Nexus need to deal with long and one-time connection. But most of the current messaging implementation are based on long connections.


packages/extension-chrome/src/background/main.ts
```diff
@@ -1,20 +1,6 @@
...
-    type: 'popup',
-    focused: true,
-    left: 200,
-    width: 360,
-    height: 600,
-    url: 'popup.html',
-  });
-});
+// listen message from content script
+onMessage('rpc', async ({ data, sender }) => server.receive(data, createRpcServerParams({ endpoint: sender })));
```
IronLu233, line 6: I think there also need a message listener for internal function invoke.
For example, import a new wallet, from mnemonic and password.

zhangyouxin (approved):
LGTM

IronLu233 (approved):

PR #41, opened 2023-01-13
feat: localforage memory driver

zhangyouxin:
# Description

We can use `localforage` to store data like `localstorage` in chrome extension. Except for all APIs in `localforage` are asynchronous. 

By default,  `localforage` support IndexedDB, WebSQL, and localstorage driver, but in test environment, we also need a memory storage driver to make `localforage` usable in node env.

# Test

- [x] Added unit tests for memory driver 

# Usage

```typescript
const memDriver = createDriver();
localforage.defineDriver(memDriver);
const instance = localforage.createInstance({ driver: 'memDriver' });

await instance.setItem("key", "value")
```


homura:
we'll use Chrome extension storage

PR #44, opened 2023-01-16
feat: split business logic into services

### LINKED ISSUE #29: service interface design and organization
For better development and testing, we need to split the Nexusinto several services, such as `WalletService`, `OwnershipService`, `ConfigServices`, etc. which will be described in TypeScript interface.

Services need to call each other, so we need a good practice to organize these services, maybe a dependency injection is required

homura:
resolved #29 

1. split business logic into several services
2. organize services and RPC server
3. a basic example of `wallet_enable` to show how to implement service and how to organize them

### DIFF
--- a/packages/extension-chrome/src/services/index.ts
+++ b/packages/extension-chrome/src/services/index.ts
@@ -1,0 +1,42 @@
+import { GrantService, NotificationService, Promisable, Storage } from '@nexus-wallet/types';
+import { createGrantService } from './grant';
+import { createInMemoryStorage } from './storage';
+import { createNotificationService } from './notification';
+
+interface Schema {
+  grant: string[];
+}
+
+export interface Services {
+  storage: Storage<Schema>;
+  grantService: GrantService;
+  notificationService: NotificationService;
+}
+
+export interface ServicesFactory {
+  get<K extends keyof Services>(name: K): Promisable<Services[K]>;
+}
+
+export function createServicesFactory(): ServicesFactory {
+  const storage = createInMemoryStorage<Schema>();
+
+  const defaultStorage = {
+    grant: [],
+  };
+  storage.setAll({
+    ...defaultStorage,
+    ...storage.getAll(),
+  });
+
+  const services = {
+    storage,
+    grantService: createGrantService({ storage }),
+    notificationService: createNotificationService(),
+  };
+
+  return {
+    get(key) {
+      return services[key];
+    },
+  };
+}
--- a/packages/extension-chrome/src/services/notification.ts
+++ b/packages/extension-chrome/src/services/notification.ts
@@ -1,0 +1,65 @@
+import { NotificationService } from '@nexus-wallet/types';
+import { errors } from '@nexus-wallet/utils';
+import browser from 'webextension-polyfill';
+
+const NOTIFICATION_WIDTH = 360;
+const NOTIFICATION_HEIGHT = 600;
+
+export function createNotificationService(): NotificationService {
+  return {
+    async requestGrant({ url }) {
+      const lastFocused = await browser.windows.getLastFocused();
+
+      const notification = await browser.windows.create({
+        type: 'popup',
+        focused: true,
+        top: lastFocused.top,
+        left: lastFocused.left! + (lastFocused.width! - 360),
+        width: NOTIFICATION_WIDTH,
+        height: NOTIFICATION_HEIGHT,
+        url: 'notification.html',
+      });
+
+      const notificationTabId = notification.tabs?.[0]?.id;
+
+      type MessageListener = Parameters<typeof browser.runtime.onMessage.addListener>[0];
+      const getRequesterAppInfoListener: MessageListener = (message, sender) =>
+        new Promise(async (sendResponse) => {
+          if (notificationTabId !== sender.tab?.id) return;
+          if (message.method !== 'getRequesterAppInfo') return;
+
+          sendResponse({ url });
+          browser.runtime.onMessage.removeListener(getRequesterAppInfoListener);
+        });
+
+      return new Promise((resolve, reject) => {
+        const userHasEnabledWalletListener: MessageListener = (message, sender) =>
+          new Promise((sendResponse) => {
+            if (notificationTabId !== sender.tab?.id) return;
+            if (message.method !== 'userHasEnabledWallet') return;
+
+            sendResponse(void 0);
+            browser.runtime.onMessage.removeListener(userHasEnabledWalletListener);
+            resolve();
+          });
+
+        browser.runtime.onMessage.addListener(getRequesterAppInfoListener);
+        browser.runtime.onMessage.addListener(userHasEnabledWalletListener);
+
+        browser.windows.onRemoved.addListener((windowId) => {
+          if (windowId === notification.id) {
+            browser.runtime.onMessage.removeListener(getRequesterAppInfoListener);
+            browser.runtime.onMessage.removeListener(userHasEnabledWalletListener);
+            reject(errors);
+          }
+        });
+      });
+    },
+    requestSignTransaction() {
+      errors.unimplemented();
+    },
+    requestSignData() {
+      errors.unimplemented();
+    },
+  };
+}
--- a/packages/types/src/services/GrantService.ts
+++ b/packages/types/src/services/GrantService.ts
@@ -1,0 +1,15 @@
+import { Promisable } from '../base';
+
+export interface GrantService {
+  /**
+   * grant the app to access the wallet
+   * @param payload
+   */
+  grant(payload: { url: string }): Promisable<void>;
+
+  /**
+   * check if an app has been granted
+   * @param payload
+   */
+  getIsGranted(payload: { url: string }): Promisable<boolean>;
+}
--- a/packages/types/src/services/KeystoreService.ts
+++ b/packages/types/src/services/KeystoreService.ts
@@ -1,10 +1,74 @@
-import { Promisable } from '../base';
+import type { BytesLike } from '@ckb-lumos/codec';
+import type { HexString } from '@ckb-lumos/lumos';
+import type { Promisable } from '../base';
+import { Provider } from './common';
 
 export interface KeystoreService {
   /**
    * check if the keystore is initialized
    */
-  isInitialized(): Promisable<boolean>;
+  hasInitialized(): Promisable<boolean>;
 
-  initKeyStore(payload: { password: string; mnemonic: string }): Promisable<void>;
+  initKeyStore(payload: InitKeyStorePayload): Promisable<void>;
+
+  /**
+   * get an extended public key by a derivation path,
+   * if the corresponding extended public is not found in store or cache,
+   * the password will be required to derive the extended public key
+   * @param payload
+   */
+  getExtendedPublicKey(payload: GetExtendedPublicKeyPayload): Promisable<string>;
+
+  /**
+   *
+   * @param payload {@link SignMessagePayload}
+   */
+  signMessage(payload: SignMessagePayload): Promisable<HexString>;
 }
+
+interface GetExtendedPublicKeyPayload {
+  path: string;
+  password?: PasswordProvider;
+}
+
+/**
+ * A {@link https://ethereum.org/en/developers/docs/data-structures-and-encoding/web3-secret-storage/ Web3 secret storage} wrapper,
+ * to interact with the keystore
+ */
+export interface InitKeyStorePayload {
+  /**
+   * password to encrypt the keystore
+   */
+  password: string;
+  /**
+   * secret item to be encrypted
+   */
+  mnemonic: string;
+
+  /**
+   * non-hardened derivation path,
+   * the corresponding public key will be stored in plain text,
+   * for example, the BIP-44 path `m / 44'/ 309'/ 0'/ 0 / 0` will store the public key of
+   * `m / 44' / 309' / 0' / 0` (external) and `m/44'/ 309'/ 0'/ 1` (internal) in plain text
+   */
+  paths: NonHardenedPath[];
+}
+
+export interface SignMessagePayload {
+  /**
+   * message be to signed
+   */
+  message: BytesLike;
+  /**
+   * derivation path of the private key, will be used to sign the message
+   */
+  path: HardenedPath | NonHardenedPath;
+  /**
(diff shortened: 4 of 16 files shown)


packages/extension-chrome/src/services/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  get<K extends keyof Services>(name: K): Promisable<Services[K]>;
+}
+
+export function createServicesFactory(): ServicesFactory {
+  const storage = createInMemoryStorage<Schema>();
+
+  const defaultStorage = {
+    grant: [],
+  };
+  storage.setAll({
```
IronLu233, line 26: Should the storage need version control?
My concern is in the future, some breaking change of persist data saving may occur.

packages/extension-chrome/src/services/notification.ts
```diff
@@ -0,0 +1,65 @@
+import { NotificationService } from '@nexus-wallet/types';
+import { errors } from '@nexus-wallet/utils';
+import browser from 'webextension-polyfill';
+
+const NOTIFICATION_WIDTH = 360;
+const NOTIFICATION_HEIGHT = 600;
+
+export function createNotificationService(): NotificationService {
```
IronLu233, line 8: can you add TODO or WIP comment here? The `requestGrant` method seems can exact common function which also be used by `requestTransaction` and `requestSigningData`

packages/types/src/services/ConfigService.ts
```diff
@@ -0,0 +1,39 @@
+import { Promisable } from '../base';
+
+export interface ConfigService {
+  getConfig(): Promisable<Config>;
+
+  updateConfig(payload: { config: Partial<Config> }): Promisable<void>;
```
IronLu233, line 6: In my opinion, the entire config getter/setter is unecessary. If I want modify specific config, for example, nickname, I will directly invoke `updateNickname`

packages/extension-chrome/src/services/index.ts
homura: right, persistent data may need to be migrated when wallet has upgraded, I'll add a version field here

packages/types/src/services/ConfigService.ts
```diff
@@ -0,0 +1,45 @@
...
+  id: string;
+  displayName: string;
+  networkName: NetworkName;
+  rpcUrl: string;
+}
+
+/**
+ * {@link https://github.com/nervosnetwork/ckb/blob/develop/rpc/README.md#type-chaininfo network type}
+ */
+export type NetworkName = 'ckb' | 'ckb_testnet' | string;
```
IronLu233, line 45: These type will be used by outside. Cloud you please export them?

packages/extension-chrome/src/services/notification.ts
homura: > can you add TODO or WIP comment here? 

marked

> The requestGrant method seems can exact common function which also be used by requestTransaction and requestSigningData

yes, if `requestTransaction` realize that the app is not granted, we can call `requestGrant` in `requestTransaction`, but this should be rare case, beacause without a grant it is not event possible to call `getUnusedLocks` or other `wallet_x` API

packages/extension-chrome/src/services/index.ts
homura: now config is versioned

packages/types/src/services/ConfigService.ts
IronLu233: I think a set/get method name pair may be more clear.

and rename `switchNetwork` to `setSelectedNetwork`.

Here is want I think
``` ts
getNetwork(): {}
setNetwork(): {}
getNickname(): {}
setNickname(): {}
getSelectedNetwork(): {}
setSelectedNetwork(): {}
```

packages/types/src/services/ConfigService.ts
```diff
@@ -0,0 +1,45 @@
...
+  updateConfig(payload: { config: Partial<Config> }): Promisable<void>;
+
+  getAvailableNetworks(): Promisable<NetworkConfig[]>;
+  updateNetwork(payload: { network: NetworkConfig }): Promisable<void>;
+  switchNetwork(payload: { id: string }): Promisable<void>;
+  removeNetwork(payload: { id: string }): Promisable<void>;
+
+  updateNickname(payload: { nickname: string }): Promisable<void>;
+
+  appendWhitelistItem(payload: { host: string }): Promisable<void>;
```
IronLu233, line 15: word `add` is just enough. like [Java's List](https://docs.oracle.com/javase/8/docs/api/java/util/List.html)

packages/types/src/services/ConfigService.ts
homura: make sense, refactored

packages/types/src/services/ConfigService.ts
homura: refactored

zhb666 (approved):

packages/types/src/services/GrantService.ts
```diff
@@ -0,0 +1,21 @@
+import { Promisable } from '../base';
+
+export interface GrantService {
+  /**
+   * grant the app to access the wallet
+   * @param payload
+   */
+  grant(payload: { host: string }): Promisable<void>;
```
zhangyouxin, line 8: The return type of `grant` has to be a `Promise`, I think narrowing down the return type might be better. `grant(payload: { host: string }): Promise<void>;`

packages/types/src/services/KeystoreService.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  mnemonic: string;
+
+  /**
+   * non-hardened derivation path,
+   * the corresponding public key will be stored in plain text,
+   * for example, the BIP-44 path `m / 44'/ 309'/ 0'/ 0 / 0` will store the public key of
+   * `m / 44' / 309' / 0' / 0` (external) and `m/44'/ 309'/ 0'/ 1` (internal) in plain text
+   */
+  paths: NonHardenedPath[];
```
zhangyouxin, line 54: Is there a scenario where user would pass in some path other than first external/change address, i.e. `m / 44'/ 309'/ 0'/ 0 / 0` when calling `initKeyStore`? or this parameter only makes it more covenient for tests.

I think when a user creates a new wallet, he would only pass in `password` and `mnemonic`,
then `getUsedLocks` in `ownershipService` would be called to detect used locks and paths.  So maybe move `paths` to `OwnershipService`

packages/types/src/services/GrantService.ts
homura: > The return type of `grant` has to be a `Promise`, I think narrowing down the return type might be better. `grant(payload: { host: string }): Promise<void>;`

the `Promisable<void>` is `Promise<void> | void`, it is convenient when we want to test some module that depend on `GrantService`


```ts

const mockedGrantService = {
  // we don't need to await here
  grant: ({ host }): void => {
    inMemoryStorage.setItem('whitelist', ...)
  },
}


function createFullOwnershipService({ grantService }) {
  return {
    signMessage() {
      if (!granted) grantService.grant({...})
    }
  }
}


// 
createOwnershipService({ grantService: mockedGrantService });

```

packages/types/src/services/KeystoreService.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export interface SignMessagePayload {
+  /**
+   * message be to signed
+   */
+  message: BytesLike;
+  /**
+   * derivation path of the private key, will be used to sign the message
+   */
+  path: HardenedPath | NonHardenedPath;
```
zhangyouxin, line 65: `HardenedPath` should not be used to sign a message.

homura: is there anything wrong with using hardened path to sign messages?

packages/types/src/services/OwnershipService.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  signData(payload: SignDataPayload): Promise<Signature>;
+}
+
+interface GetPaginateItemsPayload {
+  cursor?: string;
+}
+
+interface GetUnusedLocksPayload {
+  change?: boolean;
```
zhangyouxin, line 33: change lock should be used only when user issues a new transaction, so unused change locks should not be used in future transactions, maybe we don't need this option here?

packages/types/src/services/GrantService.ts
zhangyouxin: OK，I get it

packages/types/src/services/KeystoreService.ts
zhangyouxin: A user of Nexus Wallet

packages/types/src/services/KeystoreService.ts
zhangyouxin: Nothing wrong with that, just in most cases the user(nexus contributor) would call `signData` with `SignDataPayload`, which is constructed of `data` and `script`, but A hardend path does'nt produce a lock.

packages/types/src/services/KeystoreService.ts
homura: I got it. A service can't be called by an end user of Nexus directly, it is just a way of organizing code.

Nexus cannot directly access the user's private key unless the user grants permission and enters the password for the WSS(Web3 Secret Storage). To derive addresses without a private key, we can use a non-hardened extended [public key](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#public-parent-key--public-child-key). For security reason, we don't encourage other modules to have direct access to the keystore or even the extended public key.

To initialize the keystore, we can use these "parent" path

- full-ownership (change: false): `m/44'/309'/0'/0`
- full-ownership (change: true): `m/44'/309'/1'/0`
- rule-based-ownership (change: false): `m/49'/309'/0'/0` 
- rule-based-ownership (change: true): `m/49'/309'/1'/0`

```ts
await keystore.initKeystore({
  paths: [`m/44'/309'/0'/0`, `m/44'/309'/1'/0`, `m/49'/309'/0'/0`, `m/49'/309'/1'/0` ]
})

// derived from       `m/44'/309'/0'/0`
keystore.getPublickey(`m/44'/309'/0'/0/0`)
keystore.getPublickey(`m/44'/309'/0'/0/1`)
keystore.getPublickey(`m/44'/309'/0'/0/3`)
```

packages/types/src/services/KeystoreService.ts
homura: maybe we can only say that we won't use a hardened key to sign the message "now". 

We don't restrict which key must be used to sign, maybe there will be a proposal to identify the user, for example, using `m / identity'`, there are some scenarios that actually require for identity, such as voting.

zhangyouxin: I get it. But in that case(using hardend path in voting Dapp),  let's say using `m/44'/309'/0'` to sign a vote， the voting dapp will know every address of user(end user of nexus wallet), the privacy of UTXO doesn't exist any more.

homura: > let's say using m/44'/309'/0' to sign a vote

The hardened key can only derive the child key from the parent private key, we can't derive the child key from `m/44'/309'/0'`'s public key

packages/types/src/services/OwnershipService.ts
zhangyouxin: According to [bip-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit), **internal chains receive only coins that come from the associated external chains.**

I think that means if the address index is `n`, then `m/44'/309'/1/n` should be used in outputs only when `m/44'/309'/0/n` is in the inputs of the transaction.

packages/types/src/services/KeystoreService.ts
zhangyouxin: Ok I get it, I just misunderstood the crypto algorithm here. 

How can the Voting Dapp know the weight of a ticket signed by `m/identity'` if the voting weight depends on the balance of signer's coin balance?

packages/types/src/services/OwnershipService.ts
homura: I guess you misunderstood here, if you say change address can only be used in output, these change coin will never be spent anymore

packages/types/src/services/KeystoreService.ts
homura: The voting app is just one example of a dApp that might require an identity mechanism, and I think we can discuss how to calculate voting weight in another issue

zhangyouxin (approved):
LGTM

packages/types/src/services/ConfigService.ts
```diff
@@ -0,0 +1,56 @@
...
+  getNetworks(): Promisable<NetworkConfig[]>;
+  addNetwork(payload: { network: NetworkConfig }): Promisable<void>;
+  removeNetwork(payload: { id: string }): Promisable<void>;
+  getSelectedNetwork(): Promisable<NetworkConfig>;
+  setSelectedNetwork(payload: { id: string }): Promisable<void>;
+
+  getNickname(): Promisable<string>;
+  setNickname(payload: { nickname: string }): Promisable<void>;
+
+  addWhitelistItem(payload: { host: string }): Promisable<void>;
```
IronLu233, line 16: a function named getWhitelist seems missing

packages/types/src/services/NotificationService.ts
```diff
@@ -0,0 +1,20 @@
...
+   * @param payload
+   */
+  requestSignTransaction(payload: { tx: Transaction }): Promise<{ password: string }>;
+
+  /**
+   * request user to approve for signing binary data,
+   * will return a password to decrypt keystore if user approved and input the correct password
+   * @param payload
+   */
+  requestSignData(payload: { data: BytesLike }): Promise<{ password: string }>;
```
IronLu233, line 17: If you means sign a binary, `requestSignBinary` is more clear than `requestSignData`. "Data" is abstract that can infer everything in computer science

homura: This method is indeed used to sign all the data, but the content of the data is in byte form, so using `signBinary` is a bit verbose

packages/types/src/services/ConfigService.ts
homura: added

IronLu233 (approved):

PR #49, opened 2023-01-28
add get used lock for ownership service

zhangyouxin:
## Description

This pr implements `getUsedLocks` of `ownershipService`.

`getUsedLocks` should scan user locks according to the logic descripbed in bip-44:
https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery

## Usage

To use the ownership service, a `NotificationService`, `KeystoreService`, `Backend` and `AddressStorageService` should be created to query transactions on the CKB chain.

```typescript
const service = createOwnershipService(notificationService, keystoreService, backend, addressStorageService);
const usedLocks = await service.getUsedLocks({});
```
The `backend` service should at least provide `hasHistory` to count transactions of given lock script on the CKB chain.

```typescript
const backend = BackendProvider.getDefaultBackend();
```

## Test

- [x] Unit test based on mocked services

### DIFF
--- a/packages/extension-chrome/__tests__/services/ownership.ts
+++ b/packages/extension-chrome/__tests__/services/ownership.ts
@@ -86,3 +86,34 @@ it('ownership#get used locks return 1st lock and 6th lock', async () => {
     objects: [locks[0], locks[5]],
   });
 });
+
+it('ownership#sign data with 1st lock', async () => {
+  const keychain = Keychain.fromSeed(shortSeed);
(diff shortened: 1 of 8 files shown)


homura:
`countTx` will probe all transactions related to a lock, we just need to check if the lock has a record transaction record, 

```ts
let hasRecord = false
for await (const tx of txCollector.collect()) {
  hasRecord = true
  break;
}
```

homura:
[`KeystoreService`](https://github.com/ckb-js/nexus/blob/services/packages/types/src/services/KeystoreService.ts) will act as a Keychain, so we don't need to depend on keychain directly

PR #75, opened 2023-02-13
feat: session messenger

homura:
This PR introduced the `SessionMessenger` which helps to communicate between extension page and background service in a session scope. 

```ts
// define a CallMap to make it type-safe
type SessionMethods = {
  ping: Call<void, 'pong'>;
}

// background.js
const service = createSessionMessenger<SessionMethods>({ 
  adapter: browserExtensionAdapter, 
  sessionId: 'session1' 
});
service.register('ping', () => 'pong');

// content-script.js
const client = createSessionMessenger<SessionMethods>({
  adapter: browserExtensionAdapter,
  sessionId: 'session1'
});
await client.send('ping') // => 'pong'
```

We can check out more cases in the [test](https://github.com/ckb-js/nexus/blob/f00bcad4c0e089190d9343f45e281aaa27e6e77f/packages/extension-chrome/__tests__/messaging/session.ts)

BTW, this PR also

- defined [SessionMethods](https://github.com/ckb-js/nexus/blob/session-messenger/packages/extension-chrome/src/services/notification.ts#L10-L27) for `NotificationService` to help it communicate with extension page
- updated lumos to the latest develop
- added a simple logger
- fixed KeystoreService test performance issue

### DIFF
--- a/packages/extension-chrome/src/services/notification.ts
+++ b/packages/extension-chrome/src/services/notification.ts
@@ -1,6 +1,30 @@
-import { NotificationService } from '@nexus-wallet/types';
+import type { NotificationService } from '@nexus-wallet/types';
 import { errors } from '@nexus-wallet/utils';
 import browser from 'webextension-polyfill';
+import { TransactionSkeletonObject } from '@ckb-lumos/helpers';
+import type { HexString, Script } from '@ckb-lumos/base';
+import { Call, createSessionMessenger } from '../messaging/session';
+import { browserExtensionAdapter } from '../messaging/adapters';
+import { nanoid } from 'nanoid';
+
+export type SessionMethods = {
+  session_getRequesterAppInfo: Call<void, { url: string; favicon: string }>;
+  session_approveEnableWallet: Call<void, void>;
+
+  /**
+   * get a ready-to-sign transaction
+   * the input/output scripts of the transaction should be highlighted if they are owned by the wallet
+   */
+  session_getUnsignedTransaction: Call<void, { tx: TransactionSkeletonObject; ownedLocks: Script[] }>;
+  session_approveSignData: Call<{ password: string }, void>;
+
+  /**
+   * get bytes to be signed, the return data should detect if it can be converted to utf8 string,
+   * if so, return the utf8 string, otherwise return the hex string
+   */
+  session_getUnsignedData: Call<void, { data: HexString | string }>;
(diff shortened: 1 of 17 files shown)


packages/extension-chrome/src/services/notification.ts
```diff
@@ -1,6 +1,30 @@
...
 import { errors } from '@nexus-wallet/utils';
 import browser from 'webextension-polyfill';
+import { TransactionSkeletonObject } from '@ckb-lumos/helpers';
+import type { HexString, Script } from '@ckb-lumos/base';
+import { Call, createSessionMessenger } from '../messaging/session';
+import { browserExtensionAdapter } from '../messaging/adapters';
+import { nanoid } from 'nanoid';
+
+export type SessionMethods = {
+  session_getRequesterAppInfo: Call<void, { url: string; favicon: string }>;
```
IronLu233, line 11: if the message is from session messenger, Why here have a `session_prefix`?

homura: just keep sampe with the RPC methods to make it easy to do full text search in the project

IronLu233: They are connected by typescript reference, because you use typescript for restrict it. I think the method name can remove `session_prefix`
IDE with Typescript language server can find it's occurrences.
image

homura: Not everyone use an IDE, some developer still use an editor

IronLu233: Every developer who want to contribute Nexus, He/she will have a developer tool, with configured environment. I use neovim for this project at most time. And I also can search by the method/type references relation.

IDE, like web storm, and editor like vim/neovim/vsocde, they all have some specific ways to make their tool use Typescript language server features.

homura: Personally, I offen use `github.dev`, and for entry function names, developer could use `wallet_*` or `session_*` to do full text search

homura: I mean, this is what I wanted when I browse the entry functions/methods

image

homura: but they are NOT events, they are actually RPC method [names](https://docs.metamask.io/guide/rpc-api.html)

IronLu233 (approved):

PR #76, opened 2023-02-13
feat(extension-chrome): ownership service

zhangyouxin:
usage:

```typescript
// given backend, keystoreService, storage, notificationService

// probe task runs regularly in the background to guarantee that
// the lockInfos in storage is up to date with the CKB chain
// and make sure there are plenty of off-chain locks to use.

ProbeTask.getInstance().run();

const { fullOwnershipService, ruleBasedOwnershipService } = createOwnershipServices({
  backend,
  keystoreService,
  notificationService,
  locksManager: new LocksManager({ storage }),
});

fullOwnershipService.getOffChainLocks({});

```


packages/extension-chrome/src/services/backend/locksManager.ts
```diff
@@ -0,0 +1,217 @@
...
+    change: number;
+  };
+};
+
+export type LocksAndPointer = {
+  details: NexusLockInfos;
+  pointers: NexusLockPointers;
+};
+
+export class LocksManager {
```
homura, line 56: 1. ONLY off-chain locks can be managed by circular. On-chain info should use the cursor, because methods like `getLiveCells` may return the same cell and cause transaction failed

2. The Circular can be made simple 

```ts
interface Circular<T> {
  readonly items: T[];

  current() :T;
  next(): T;
  takeAndSupply(take: number | (item: T) => boolean, newScript: T): T;
}

// off-chain locks provider
let fullInternalProvider: Circular<LockInfo>;
let fullExternalProvider: Circular<LockInfo>;
let ruleBasedProvider: Circular<LockInfo>;

// getOffChainLocks
return fullExtenalProvider.next();

// mark as on-chan lock
fullExternalProvider.takeAndSupply((info) => {
  return info.lockHash === '0x...'
}, derived)

```

packages/extension-chrome/src/services/ownershipService.ts
```diff
@@ -0,0 +1,96 @@
...
+          password,
+          path: lockInfo.path,
+        });
+        result.push([lockInfo.lock, signature]);
+      }
+      return result;
+    },
+    signData: async (payload: SignDataPayload) => {
+      const lockInfo = config.locksProvider.getlockInfoByLock({ lock: payload.lock });
+      asserts.nonEmpty(lockInfo);
```
homura, line 86:
```ts
asserts(
  lockInfo, 
  "payload.lock is not derived by the wallet, please check if the lock is correct", 
  payload.lock,
)
```

We should make the error messages clear and understandable by the dApp developer.

homura:
I reviewed the test cases related to full ownership, but to be honest, the completion of this PR is still [too low](https://app.codecov.io/gh/ckb-js/nexus/commit/5271f0389d1054a08a20baa1fdb8e574e1b75040/blob/packages/extension-chrome/src/services/ownershipService.ts)

- no paginated test case of `getLiveCells`
- no matched test case of `getOffChainLocks`, only with a `toHaveLength(1)`
- no non-empty test case of `getOnCHainLocks`
- no non-empty test case of `signTransaction`
- no non-empty test case of `signData`

these cases are probably the most important of this PR for a reviewer. As a reviewer, if there are some cases like this, it would be more friendly

```ts
describe('getLiveCells', async () => {
  it('should only the first page of cells will be fetched', () => {
    mockedBackend.pushCells([createCell({lock: key1}, createCell({lock: key2}))])
    probeTask.start();
    await sleep(100);
    expect(ownership.getLiveCells()).resolve.toMatchObject([
      createCell({lock: key1}, 
      createCell({lock: key2}))
    ])
  });

  it('should paginated when the cursor is non-empty', () => {
    const cells = [withLock1, withLock2, ...,  withLock15] 
    ...
    expect(ownership.getLiveCells()).resolve.toMatchObject(first10Cells);
    expect(ownership.getLiveCells()).resolve.toMatchObject(last5Cells)
  })
})
```

zhangyouxin:
> I reviewed the test cases related to full ownership, but to be honest, the completion of this PR is still [too low](https://app.codecov.io/gh/ckb-js/nexus/commit/5271f0389d1054a08a20baa1fdb8e574e1b75040/blob/packages/extension-chrome/src/services/ownershipService.ts)
> 
> * no paginated test case of `getLiveCells`
> * no matched test case of `getOffChainLocks`, only with a `toHaveLength(1)`
> * no non-empty test case of `getOnCHainLocks`
> * no non-empty test case of `signTransaction`
> * no non-empty test case of `signData`
> 
> these cases are probably the most important of this PR for a reviewer. As a reviewer, if there are some cases like this, it would be more friendly
> 
> ```ts
> describe('getLiveCells', async () => {
>   it('should only the first page of cells will be fetched', () => {
>     mockedBackend.pushCells([createCell({lock: key1}, createCell({lock: key2}))])
>     probeTask.start();
>     await sleep(100);
>     expect(ownership.getLiveCells()).resolve.toMatchObject([
>       createCell({lock: key1}, 
>       createCell({lock: key2}))
>     ])
>   });
> 
>   it('should paginated when the cursor is non-empty', () => {
>     const cells = [withLock1, withLock2, ...,  withLock15] 
>     ...
>     expect(ownership.getLiveCells()).resolve.toMatchObject(first10Cells);
>     expect(ownership.getLiveCells({cursor: page2})).resolve.toMatchObject(last5Cells)
>   })
> })
> ```

added test cases mentioned above, and now the branch coverage is [higher](https://app.codecov.io/gh/ckb-js/nexus/commit/cc84f605a37c62d982b55ff61145a76f7f8301f8/blob/packages/extension-chrome/src/services/ownershipService.ts)

packages/extension-chrome/__tests__/services/ownership/ownership.ts
```diff
@@ -0,0 +1,564 @@
...
+    jest.clearAllMocks();
+  });
+  it('should get next offchain locks', async () => {
+    expect(await fullOwnershipService.getOffChainLocks({})).toEqual([
+      {
+        args: '0x',
+        codeHash: '0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8',
+        hashType: 'type',
+      },
+    ]);
```
homura, line 262: We should use different locks in these cases, otherwise we can't know that the next lock is really next, because they are all the same

zhangyouxin: updated

packages/extension-chrome/src/services/ownership/ruleBased/utils.ts
```diff
@@ -0,0 +1,37 @@
+import { LockInfoStorage, RuleBasedLocksAndPointer } from '../types';
+import { Storage } from '@nexus-wallet/types';
+
+export const RULE_BASED_HARDENED_PATH = "m/4410179'/0'";
```
homura, line 4: it should not be a hardened path because a hardened public key cannot derive child key, we should change the design to `m/ckb_purpose 1'/account'/change(always 0)/index`

packages/extension-chrome/src/services/ownershipService.ts
```diff
@@ -0,0 +1,185 @@
...
+        });
+        result.push([lockInfo.lock, signature]);
+      }
+      return result;
+    },
+    signData: async (payload: SignDataPayload) => {
+      const lockInfo = await config.locksManager.getlockInfoByLock({ lock: payload.lock });
+      asserts.asserts(lockInfo, 'Lock not found when call signData with', payload);
+      const password = (await config.notificationService.requestSignData({ data: payload.data })).password;
+      const signature = await config.keystoreService.signMessage({
```
homura, line 147: to avoid security risk, the signData must with some [magick bytes]( https://github.com/nervosnetwork/neuron/blob/c46fe259abf427524866329f98512f251a5e7c07/packages/neuron-wallet/src/services/sign-message.ts#L14), we can mark a TODO here

packages/extension-chrome/src/rpc/server.ts
```diff
@@ -17,6 +19,20 @@ export function addMethod<K extends keyof (WalletMethods & DebugMethods)>(
 }
 
 let servicesFactory = createServicesFactory();
+void probeRun();
+
+async function probeRun() {
+  const keystoreService = createServicesFactory().get('keystoreService');
+  const storage = createServicesFactory().get('storage');
+  if (!keystoreService || !storage) return;
+  console.log('wallet already initialized, start probe task...');
```
homura, line 28: we can use the logger instead of `console.log`

packages/extension-chrome/src/rpc/server.ts
```diff
@@ -17,6 +19,20 @@ export function addMethod<K extends keyof (WalletMethods & DebugMethods)>(
 }
 
 let servicesFactory = createServicesFactory();
+void probeRun();
+
+async function probeRun() {
+  const keystoreService = createServicesFactory().get('keystoreService');
+  const storage = createServicesFactory().get('storage');
```
homura, line 26: refactor the `keystoreSertvice` and `storage` to params instead of `createServiceFactory` everywhere

packages/extension-chrome/src/rpc/types.ts
```diff
@@ -1,13 +1,25 @@
...
+import { Call, CallMap, Paginate } from '@nexus-wallet/types';
+import { Signature, GroupedSignature } from '@nexus-wallet/types/lib/injected';
 
 export interface WalletMethods extends CallMap {
   wallet_enable: Call<void, void>;
   wallet_isEnabled: Call<void, boolean>;
   wallet_getNetworkName: Call<void, string>;
 
-  wallet_fullOwnership_getUnusedLocks: Call<void, Script[]>;
+  wallet_fullOwnership_getOffChainLocks: Call<GetOffChainLocksPayload, Promise<Script[]>>;
```
homura, line 18: `Promise` is unnecessary in `Call`

packages/extension-chrome/src/rpc/walletImpl.ts
```diff
@@ -24,15 +30,121 @@ addMethod('wallet_enable', async (_, { getRequesterAppInfo, resolveService }) =>
   }
 
   await configService.addWhitelistItem({ host: host, favicon: `${protocol}//${host}/favicon.ico` });
+  const servicesFactory = createServicesFactory();
```
homura, line 33: use `resolveService` instead of createServicesFactory() everywhere

packages/extension-chrome/src/rpc/walletImpl.ts
```diff
@@ -24,15 +30,121 @@ addMethod('wallet_enable', async (_, { getRequesterAppInfo, resolveService }) =>
...
+    keystoreService,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    storage: storage as any,
+  });
+
+  probe.run();
+});
+
+addMethod('wallet_fullOwnership_getOffChainLocks', async (payload) => {
+  console.log('====================================');
```
homura, line 49: `======` is unnecessary, if you just want to make a different logger, please createLogger('special_module_name') to instead

packages/extension-chrome/src/rpc/walletImpl.ts
```diff
@@ -24,15 +30,121 @@ addMethod('wallet_enable', async (_, { getRequesterAppInfo, resolveService }) =>
...
+  console.log('wallet_fullOwnership_getOffChainLocks');
+  const servicesFactory = createServicesFactory();
+  const keystoreService = await servicesFactory.get('keystoreService');
+  const notificationService = await servicesFactory.get('notificationService');
+  const storage = await servicesFactory.get('storage');
+  console.log('storage', await storage.getItem('full' as never));
```
homura, line 55: avoid using `never` as a parameter, it is not [recommendded](https://www.typescriptlang.org/docs/handbook/2/functions.html#never)

homura:
This PR has many issues, and we should plan to reimplement this PR 

## Complex Asynchronous

Ownership need to handle many async, this PR uses `Promise.all` and `for + Promise` to implement async. When we need to control the concurrency of async, this would be a bad solution. For example, when we need to probe if the next 100 on-chain locks have cells or not, `Promise.all` may cause the server to reject too many connection requests

## Not Enough Abstraction

The `full-external` `full-internal` `rule-based` are similar, and we don't need to treat them differently, we can just use compose instead of the cross these condition like `getRuleBasedOffChainLocks`

```ts
interface DerivedLocks {
  lockInfos: LockInfo[];
}

const isFullOwnership = (info) => info.parantPath === FULL_OWNERSHIP_PARENT_PATH;
const isRuleBasedOwnership = (info) => info.parentPath === RULE_BASED_OWNERSHIP_PARENT_PATH;
const isOnChain = (info) => info.change === 'external';
const ifOffChain = (info) => info.change === 'internal';

// an async db driver
const infos = await lockInfos.filter(compose(isFullOwnership, isOffChain));
...

```

Therefor, we will not merge this PR, but we will use this branch for writing e2e cases

PR #78, opened 2023-02-16
feat: internal_initWallet for init wallet

homura:
## `internal_initWallet` 

To init wallet in the extension page

```ts
// Init.tsx
const factory = createServicesFactory()
const internalService = await factory.get('internalService');
internalService.initWallet({
  mnemonic: 'bandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about',
  password: '123456',
  nickname: 'Alice',
})
```

## `debug_initWallet` 

To init wallet when developing

```ts
// browser console

// will init the wallet with
// password: 123456
// mnemonic: abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about
await ckb.request({ method: 'debug_initWallet' });

const provider = await ckb.enable()
provider.fullOwnership...
```

## Adapt to Extension Storage

This PR use the browser extension local storage as the database. There is no UI for exploring stored data, if we want to explore or clear data, we can use `await chrome.storage.local.get()` or `await chrome.storage.local.clear()` in the Nexus's devtools


.eslintrc.js
```diff
@@ -16,6 +21,7 @@ module.exports = {
       rules: {
         '@typescript-eslint/no-explicit-any': 'error',
         '@typescript-eslint/explicit-module-boundary-types': 'error',
+        '@typescript-eslint/no-floating-promises': 'error',
```
IronLu233, line 24: I don't vote for adding this rule as error. because in many case we invoke a async function but do not handle the resolving. If floating promises as an error, we have to add `void` explicitly. a warning may be better.

homura: yup, this rule force us to add a `void` for promises without an `await`, warn is not enough since we only want CI check the error lint

```ts
const result = fetch('/something') ❌
const result = await fetch('/something') ✅

fetch('/something') ❌
doOtherThing()

void fetch('/something') ✅
doOtherThing()
```

we can check the [diff](https://github.com/ckb-js/nexus/pull/78/files#diff-3e98cc4d19fedfce432088bbcbd323e48800a0c7cceafc8e0772243cecb59dd8L13-L14). if we don't await the navigation, and we don't have re-fetch in `/network`, it may still displaying the expired content

IronLu233 (approved):

PR #84, opened 2023-02-22
feat(extension-chrome): Add signing data implement

IronLu233:
This PR implement the sign message UI logic

when call `NotificationService#requestSignData`, it will open a notification dialog.

image

then user input the correspond  password, the method will resolved and return the wallet password.


packages/extension-chrome/src/contentScript/inpage.ts
```diff
@@ -29,7 +30,13 @@ const injectedCkb: InjectedCkb = {
...
 
         async signData(payload) {
-          return client.request('wallet_fullOwnership_signData', payload);
+          let { data } = payload;
+          if (typeof data === 'string') {
+            data = /^0x([0-9a-fA-F][0-9a-fA-F])+$/i.test(data) ? data : bytes.hexify(bytes.bytifyRawString(data));
+          } else {
+            data = bytes.hexify(data);
+          }
+          return client.request('wallet_fullOwnership_signData', { ...payload, data });
```
homura, line 39: `bytifyRawString` can't use here, the `bytifyRawString` only allowed ASCII characters, `client.request('wallet_fullOwnership_signData')` is fine

packages/extension-chrome/src/pages/Notification/containers/SignData.tsx
```diff
@@ -1,51 +1,135 @@
...
+    [checkPassword],
+  );
+
+  const dataForSigning = useMemo(() => {
+    if (!unsignedDataQuery.data) return '';
+    const unsigned = bytes.bytify(unsignedDataQuery.data.data);
+
+    if (isUtf8(unsigned)) {
+      const decoder = new TextDecoder('utf-8');
+      return decoder.decode(new Uint8Array(unsigned));
```
homura, line 70: `decoder.decode(unsigned)` is fine

IronLu233: image
Arbitrary data will parsed to garbled characters. So it is necessary.

packages/extension-chrome/src/pages/Notification/containers/SignData.tsx
```diff
@@ -1,51 +1,135 @@
...
+          />
+          <FormErrorMessage>Password Incorrect!</FormErrorMessage>
+        </FormControl>
+
+        <ButtonGroup mt="32px" size="md">
+          <Button isLoading={formState.isSubmitting} onClick={onReject} w="220px" color="gray.800" colorScheme="gray">
+            Reject
+          </Button>
+
+          <Button w="220px" isLoading={formState.isSubmitting} type="submit">
```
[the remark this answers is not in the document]
IronLu233: Not necessary, https://github.com/ckb-js/nexus/pull/84/files#diff-7655d6674f9840d6114339579e9206b751ca35efe5b9836731c4066f5bdde52cR82 singleton can ensure the data is accessable.

packages/extension-chrome/src/pages/Notification/containers/SignData.tsx
homura: the unsigned is alredy a `Uint8Array`, this line of code is equivalent to `new Uint8Array(new Uint8Array(something))`

IronLu233: But is it possible that the unsigned is a binary data?

packages/extension-chrome/src/contentScript/inpage.ts
IronLu233: But I met a problem if the user pass Typed array, the JSONRPC will parse it as a object. I think a better way is: if it's typed array, parse it to hex string, otherwise, as origin.

homura: I mean, `bytifyRawString` will cause bugs, it has nothing to do with `TypedArray`

JSON-RPC already specifies the use of JSON as the communication protocol, `TypedArray` can not be supported in the first place, if you want to support `TypedArray` we can discuss in another place

homura: I might understand what the problem is here, currently `CkbProvider` specifies `signData(payload: { data: ByteLike }): Promise<Signature>`, the modification here tries to support all characters, but `bytifyRawString('👋')` throws an error while `bytifyRawString('Hi')` does not, so maybe we should think of `inpage.js` not as an SDK, but as a pure JSON-RPC client

If we want to support other characters, we should provide them through a library like ethers.js or web3.js. `inpage.js` should be kept as simple as possible

homura:
image

There is an error while compiling, why we should change the requestSignData's payload from `BytesLike` to `ArrayBufferLike`

packages/extension-chrome/src/pages/Notification/containers/SignData.tsx
```diff
@@ -1,51 +1,129 @@
...
+            })}
+            background="white"
+            color="black"
+            data-test-id="password"
+          />
+          <FormErrorMessage>Password Incorrect!</FormErrorMessage>
+        </FormControl>
+
+        <ButtonGroup mt="32px" size="md">
+          <Button isLoading={formState.isSubmitting} onClick={onReject} w="220px" color="gray.800" colorScheme="gray">
```
homura, line 118:
```suggestion
          <Button isDisabled={formState.isSubmitting} onClick={onReject} w="220px" color="gray.800" colorScheme="gray">
```

homura (approved):

PR #89, opened 2023-02-27
feat: implement ownership

homura:
this PR is inspired by https://github.com/ckb-js/nexus/pull/76, but with some refactor

- added an auto-incremental id as the local cursor in script info db
- batched backend to avoid network IO performance issue
- supporting multiple networks
- readability optimization


packages/extension-chrome/src/services/ownership/fullOwnership.ts
```diff
@@ -0,0 +1,138 @@
...
+    return createScriptInfoDb({ storage, networkId: selectedNetwork.id });
+  }
+
+  async function getLumosConfig(): Promise<Config> {
+    const backend = await backendProvider.resolve();
+
+    return {
+      PREFIX: 'ckb',
+      SCRIPTS: { SECP256K1_BLAKE160: await backend.getSecp256k1Blake160ScriptConfig() },
+    } satisfies Config;
```
zhangyouxin, line 35: nice usage of new feature in ts4.9

packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -0,0 +1,70 @@
...
+import { createTransactionSkeleton, LiveCellFetcher, TransactionSkeletonType } from '@ckb-lumos/helpers';
+import { throwError } from '@nexus-wallet/utils/lib/error';
+import { ScriptConfig } from '@ckb-lumos/config-manager';
+
+export interface Backend {
+  getSecp256k1Blake160ScriptConfig(): Promise<ScriptConfig>;
+
+  hasHistories(payload: { locks: Script[] }): Promise<boolean[]>;
+
+  getLiveCellsByLocks(payload: {
```
zhangyouxin, line 15: Not so easy to implement this interface here. The `cursor` may belong to some lock in the `locks` passed in here, but we don't know which lock it belongs to.

homura: the comment is not so clearly here, the cursor here must be the first lock's, to implement it

```ts
const [lock, ...otherLocks] = locks;

const thunkedRequest = [ 
//     the cursor must belong to this lock
//           👇
  [{ lock, cursor }, {lock: otherLocks[1]}, {lock: otherLocks[2]}, ...]
  [{ lock: otherLocks[thunkLength - 1] }, ...]
  [...] 
]
const thunk0 = await rpc.getHistory(thunkedRequest[0])
...
const thunkn = await rpc.getHistory(thunkedRequest[n])


return { objects:[...], cursor: thunkn[-1].cursor }
```

zhangyouxin: OK, that's reasonable

zhangyouxin (approved):
LGTM

PR #94, opened 2023-02-28
feat(extension-chrome): fullownership backend & rpc

zhangyouxin:
This PR implements full ownership backend service and rpc methods.

Todo:

- [WIP] add test cases to ownership service

### DIFF
--- a/packages/extension-chrome/src/background/main.ts
+++ b/packages/extension-chrome/src/background/main.ts
@@ -4,7 +4,9 @@ import '../rpc/debugImpl';
 import { createLogger, LIB_VERSION } from '@nexus-wallet/utils';
 import { Endpoint, onMessage } from 'webext-bridge';
 import { createServer } from '../rpc';
-import { makeBrowserExtensionModulesFactory } from '../services';
+import { makeBrowserExtensionModulesFactory, ModulesFactory } from '../services';
+import { createWatchtower } from '../services/ownership';
+import { createScriptInfoDb, OwnershipStorage } from '../services/ownership/storage';
 
 const logger = createLogger();
 logger.info(`Hi, this is Nexus@%s`, LIB_VERSION);
@@ -14,6 +16,7 @@ if (process.env.NODE_ENV === 'development') {
 
 const factory = makeBrowserExtensionModulesFactory();
 const server = createServer<Endpoint>(factory);
+void runWatchtower(factory);
 
 // listen message from content script
 onMessage('rpc', async ({ data, sender }) => {
@@ -26,3 +29,16 @@ onMessage('rpc', async ({ data, sender }) => {
     throw error;
   }
 });
+
+async function runWatchtower(factory: ModulesFactory) {
+  const configService = factory.get('configService');
+  const keystoreService = factory.get('keystoreService');
+  const backendProvider = factory.get('backendProvider');
+  const storage = factory.get('storage') as OwnershipStorage;
+
+  const selectedNetwork = await configService.getSelectedNetwork();
+  const db = createScriptInfoDb({ storage, networkId: selectedNetwork.id });
+
+  const watchtower = createWatchtower({ db, keystoreService, backend: await backendProvider.resolve() });
+  watchtower.run();
+}
--- a/packages/extension-chrome/src/services/ownership/backend.ts
+++ b/packages/extension-chrome/src/services/ownership/backend.ts
@@ -1,11 +1,11 @@
 import type { Cell, HashType, HexString, Script, Transaction } from '@ckb-lumos/lumos';
 import { RPC } from '@ckb-lumos/lumos';
 import type { ConfigService, Paginate, Promisable } from '@nexus-wallet/types';
-import { errors } from '@nexus-wallet/utils';
+import { asserts } from '@nexus-wallet/utils';
 import { NetworkId } from './storage';
 import { createTransactionSkeleton, LiveCellFetcher, TransactionSkeletonType } from '@ckb-lumos/helpers';
 import { throwError } from '@nexus-wallet/utils/lib/error';
-import { ScriptConfig } from '@ckb-lumos/config-manager';
+import { ScriptConfig, getConfig } from '@ckb-lumos/config-manager';
 
 export interface Backend {
   getSecp256k1Blake160ScriptConfig(): Promise<ScriptConfig>;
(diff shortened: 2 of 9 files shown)


packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -31,24 +30,208 @@ const _Secp256k1Blake160ScriptInfoCache = new Map<NetworkId, ScriptTemplate>();
...
+  // const rpc = new RPC(_payload.rpc);
 
   return {
-    getSecp256k1Blake160ScriptConfig: () => {
-      errors.unimplemented();
+    getSecp256k1Blake160ScriptConfig: async (): Promise<ScriptConfig> => {
+      const lumosConfig = getConfig();
+      const secp256k1Config = lumosConfig.SCRIPTS['SECP256K1_BLAKE160'];
+      asserts.asserts(secp256k1Config, 'SECP256K1_BLAKE160 script config not found');
+      return Promise.resolve(secp256k1Config);
```
homura, line 41: this may don't work with local network, we can refer to ckit's [implementation](https://github.com/ckb-js/ckit/blob/342896f5e7c0ab82a9f78ebd7ca73d9f0ec70876/packages/ckit/src/__tests__/deploy.ts#L27)

packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -31,24 +30,208 @@ const _Secp256k1Blake160ScriptInfoCache = new Map<NetworkId, ScriptTemplate>();
...
-    getLiveCellsByLocks: () => {
-      errors.unimplemented();
+    getLiveCellsByLocks: async (payload: {
+      locks: Script[];
+      /**
+       * the cursor is the ckb indexer cursor
+       */
+      cursor?: string;
+      // TODO implement limit
+      limit?: number;
```
homura, line 82: we need a default limit here, because if not, all the cells will be returned at once, and when there are too many cells, it will crash the wallet

packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -31,24 +30,208 @@ const _Secp256k1Blake160ScriptInfoCache = new Map<NetworkId, ScriptTemplate>();
...
+          },
+          body: JSON.stringify(requestParam),
+          method: 'POST',
+        });
+        const responses = await rawResponse.json();
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return responses.map((response: any) => toPaginatedCells(response));
+      };
+
+      const chunkedRequests = [];
```
homura, line 153: please add test case for this part or replace it with `lodash.chunk`

packages/extension-chrome/src/services/ownership/backend.ts
zhangyouxin: updated

packages/extension-chrome/src/services/ownership/backend.ts
zhangyouxin: setting it to 20 by default

packages/extension-chrome/src/services/ownership/backend.ts
zhangyouxin: replaced with `lodash.chunk`, then tested it manually afterwards.

packages/extension-chrome/src/services/ownership/backend.ts
homura: if it defaults to 20, we should check the cursor from the indexer, because of the cursor may not the cursor of the last cell from `getLiveCellsByLocks()`

```ts
// batched result
function getLiveCells() {
  const batchedResultFromIndexer = [cell1, cell2, cell3]
  const indexerCursor = cell3Cursor

  return {
    // return 2 cells since limit is 2
    objects: [cell1, cell2], 
    cursor: indexerCursor // ❌ should be cell2Cursor here, but it is cell3Cursor
  }
}

getLiveCells({ limit: 2 })
 
```

To fix it, we should fetch it again from the indexer with 
```js
const fetched = indexer.request({
  // desc from the last cursor
  order: "desc",
  limit: lastIndexedCells.length - limit // 3 - 2 [cell1, cell2, cell3] - [cell1, cell2] -> cell3's cursor
})

// check the desc fetch actually we wanted
asserts(fetched[-1].outPoint  == cell2.outPoint)
```

packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -31,24 +31,216 @@ const _Secp256k1Blake160ScriptInfoCache = new Map<NetworkId, ScriptTemplate>();
...
+        jsonrpc: '2.0',
+        method: 'get_cells',
+        params: [
+          {
+            script: {
+              code_hash: lock.codeHash,
+              hash_type: lock.hashType,
+              args: lock.args,
+            },
+            script_type: 'lock',
```
homura, line 102:
```suggestion
            script_type: 'lock',
            script_search_mode: 'exact',
```

the indexer defaults to prefix match, we should use the [exact](https://github.com/nervosnetwork/ckb-indexer#get_cells) match here to avoid unexpected cells from the indexer

packages/extension-chrome/src/services/ownership/backend.ts
zhangyouxin: updated and tested manually

packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -177,21 +178,70 @@ export function createBackend(_payload: { rpc: string }): Backend {
...
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(descSearchParam),
+          method: 'POST',
+        });
+        const descSearchResult = toPaginatedCells(descSearchResp.json());
+        asserts.asserts(
+          isEqual(descSearchResult.objects[-1], result.objects[limit - 1]),
```
homura, line 237: dose the current typescript supported array[-1] syntax? The above code is just pseudocode

packages/extension-chrome/src/services/ownership/backend.ts
zhangyouxin: updated in `get_transactions` too

packages/extension-chrome/src/services/ownership/backend.ts
zhangyouxin: No, it will always returns `undefined`, not a scenario on the chain, so this assert never runs. I find it necessary to collect some fixtures and test `backend.ts`, plan to do it next 

The code is updated.

packages/extension-chrome/__tests__/services/ownership/backend.test.ts
```diff
@@ -0,0 +1,279 @@
...
+function getSimpleScenario() {
+  const scenario: Record<number, number> = {
+    0: 22,
+  };
+  const locks = [createLock(0)];
+  return { scenario, scenarioAnswer: '0x00:0x13', locks };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const getMockFetchLiveCellsResp = (scenario: any) => (req: Request) => {
```
homura, line 201:
```suggestion
type LockIndex = number;
type FetchedCellCount = number;
type Scenario = Record<LockIndex, FetchedCellCount>;

const getMockFetchLiveCellsResp = (scenario: Scenario) => (req: Request) => {
```

packages/extension-chrome/src/background/main.ts
```diff
@@ -38,3 +41,18 @@ onMessage('rpc', async ({ data, sender }) => {
     throw error;
   }
 });
+
+async function runWatchtower(factory: ModulesFactory) {
```
homura, line 45:
```suggestion
// TODO: listen for `networkChanged` event and restart the watchtower
async function runWatchtower(factory: ModulesFactory) {
```

packages/extension-chrome/src/services/ownership/backend.ts
```diff
@@ -29,26 +30,270 @@ type ScriptTemplate = { codeHash: HexString; hashType: HashType };
...
+export function createBackend(_payload: { nodeUrl: string }): Backend {
+  // TODO use rpc to fetch onchain data when lumos rpc is ready to use in chrome extension
+  const rpc = new RPC(_payload.nodeUrl);
 
   return {
-    getSecp256k1Blake160ScriptConfig: () => {
-      errors.unimplemented();
+    getSecp256k1Blake160ScriptConfig: async (): Promise<ScriptConfig> => {
+      const lumosConfig = getConfig();
+      let secp256k1Config = lumosConfig.SCRIPTS['SECP256K1_BLAKE160'];
```
homura, line 40: we should NOT use the config from Lumos since we don't know what network is connected, it is recommended to implement it via

```ts
const Secp256k1Blake160ScriptInfoCache = new Map<NetworkId, ScriptTemplate>();

return {
  getSecp256k1Blake160ScriptConfig: () => {
    if (!Secp256k1Blake160ScriptInfoCachehas(selectedNetworkId)) {
      Secp256k1Blake160ScriptInfoCachehas.set(
        selectedNetworkId, 
        await loadSecp256k1ScriptDep()
      )
    }

    return Secp256k1Blake160ScriptInfoCache.get(selectedNetworkId)!;
    
  }
}

```

homura:
I have tried this PR on https://demo-nexus.vercel.app, however after I received my CKB from faucet, the balance is always 0

packages/extension-chrome/__tests__/services/ownership/backend.test.ts
```diff
@@ -0,0 +1,307 @@
+import { Script, Transaction } from '@ckb-lumos/lumos';
+import { createBackend } from '../../../src/services/ownership/backend';
+import fetchMock from 'jest-fetch-mock';
+
+describe('hasHistory', () => {
+  afterEach(() => {
+    fetchMock.resetMocks();
+  });
+  it('should get history by lock list', async () => {
+    fetchMock.mockResponse(mockFetchTransactionsResp);
```
homura, line 10: A small suggestion, if there is only one mock, then we can forgive to keep it simple and not need a separate function

```js
fetchMock.mockResponse(JSON.stringify([
  { last_cursor: '', objects: [] },
  { last_cursor: 'tx_cursor', objects: [{} /* Transaction */] },
  { last_cursor: '', objects: [] },
]))
```

packages/extension-chrome/__tests__/services/ownership/backend.test.ts
```diff
@@ -0,0 +1,307 @@
...
+              hash_type: lock.hashType,
+              args: lock.args,
+            },
+            type: null,
+          },
+        },
+        status: 'live',
+      },
+      id: 1,
+    }),
```
homura, line 305: `createJSONRPCSuccessResponse` from `json-rpc-2.0` is fine

homura (approved):

PR #96, opened 2023-03-01
feat(extension-chrome): add network and whitelist sites implementation

IronLu233:
# This PR takes full implemented freature of wallet whitelist and network select. Includes these features:
- Search and remove whitelist sites
- toggle network of blockchain
- add new network of blockchain

# Screenshots
image
image


packages/extension-chrome/src/pages/Popup/containers/Network/NetworkConfig.tsx
```diff
@@ -1,44 +1,33 @@
...
-        configService.getSelectedNetwork(),
-      ]);
-      return {
-        networks,
-        selectedNetwork,
-      };
+  const configQuery = useConfigQuery();
+  const toggleNetworkMutation = useMutation({
+    mutationFn: (id: string) => {
+      const configService = createServicesFactory().get('configService');
```
homura, line 15: although it will work, it is better to have only one factory in a process, maybe we can make a hook to resolve Nexus modules

```tsx
// index.tsx
function main() {
  return (
    <NexusModulesProvider>
      <Route ... />
    </NexusModulesProvider>
  )
}

// hooks/useNexusModules.ts
declare function useNexusModules<K extends keyof Modules>(name: K): Modules[K];

// some-page.tsx

function SomePage() {
  const configService = useNexusModules('configService');
}

```

IronLu233: I think the context can be omitted, because the lifecycle is managed by awilix.
``` ts
const useConfigService<T extends keyof Module>(name: T): Module[T] {
  return createServicesFactory().get(name)
}
```

homura: every `createServiceFactory()` will make a new aliwix container, now the modules in the factory are stateless, so you can't sense any changes here, but if a module in the factory becomes a stateful module, the singleton module will have a trouble

```ts
createServicesFactory().get(name) === createServicesFactory().get(name) // false 
factory.get(name) === factory.get(name) // true
```

IronLu233: https://github.com/ckb-js/nexus/pull/96/files#diff-6e6e1774d88c5c03bf0ae25dd00334e0f6bccf5196fd66b63d35294aa2351c15R1

Finally, I make it as a global singleton, and can use it via`useService`

packages/extension-chrome/src/pages/Popup/containers/Network/AddNetwork.tsx
```diff
@@ -1,27 +1,39 @@
...
-import { WhiteAlphaBox } from '../../../Components/WhiteAlphaBox';
 import { AddIcon } from '@chakra-ui/icons';
 import { useForm } from 'react-hook-form';
 import { nanoid } from 'nanoid';
+import { useMutation } from '@tanstack/react-query';
+
+import { WhiteAlphaBox } from '../../../Components/WhiteAlphaBox';
+import { useService } from '../../../hooks/useService';
+
+const HTTP_URL_PATTERN = /https?:\/\/[a-zA-Z_\-.~]+/;
```
homura, line 12: it seems failed when match `http://localhost:3000`

packages/extension-chrome/src/pages/Popup/containers/WhitelistSites.tsx
```diff
@@ -10,38 +10,46 @@ import {
...
-  const removeSite = (site: string) => async () => {
-    await removeWhitelistSiteMutation.mutateAsync(site);
-    await whitelistSiteQuery.refetch();
+  const removeSite = (host: string) => async () => {
+    try {
+      await removeWhitelistItemMutation.mutateAsync(host);
+    } catch {
+      toast({ title: "Can't remove the site" });
+    } finally {
+      await configQuery.refetch();
```
homura, line 40: it is better to invalid `configQuery`'s cache after mutated config instead of refetch it because there may be more than one place where `configService.getConfig` is used

packages/extension-chrome/src/pages/Popup/containers/Network/AddNetwork.tsx
IronLu233: It works. But has other issue. I have optimized it.

homura: i find the new regex also mismatched with `https://testnet.ckb.dev/rpc`, maybe `(https?://.*):(\d*)\/?(.*)` is fine

IronLu233: So I add `/` to match set. Here they are all matched
image
image

packages/extension-chrome/src/pages/Popup/containers/Network/AddNetwork.tsx
```diff
@@ -1,29 +1,41 @@
...
+    e?.preventDefault();
+    await addNetworkMutation.mutateAsync({ name, rpcUrl });
     navigate(-1);
   });
 
+  const addNetworkMutation = useMutation({
+    mutationFn: ({ name, rpcUrl }: AddNetworkFormState) => {
+      return configService.addNetwork({
+        network: { displayName: name, rpcUrl, id: nanoid(), networkName: name },
+      }) as Promise<void>;
```
homura, line 35:
```ts
      return Promise.resolve(configService.addNetwork({
        network: { displayName: name, rpcUrl, id: nanoid(), networkName: name },
      }));
```

`Promise.resolve` is fine, we should avoid using the `as` whenever possible

IronLu233: The best way I think is add async on the function signature.
```
async ({ name, rpcUrl }: AddNetworkFormState) => {
      return configService.addNetwork({
        network: { displayName: name, rpcUrl, id: nanoid(), networkName: name },
      })
```

IronLu233:
@homura Can you help confirm these unresolved conversation?

packages/extension-chrome/src/pages/hooks/useConfigQuery.ts
```diff
@@ -0,0 +1,16 @@
+import { Config } from '@nexus-wallet/types/lib/services';
+import { useQuery, useQueryClient, UseQueryResult } from '@tanstack/react-query';
+import { useService } from './useService';
+
+export function useConfigQuery(): UseQueryResult<Config, unknown> & { invalidate: () => Promise<void> } {
+  const configService = useService('configService');
+  const queryClient = useQueryClient();
+
+  return {
+    ...useQuery({
```
homura, line 10: have you tested it? If `useQuery` return an instance of a class, the `...` operator may cause `this` to lose the context

IronLu233: It works well. I test it. I will confirm it in the source code soon

IronLu233: without the explicit affects, I prevent this implementation.

packages/extension-chrome/src/pages/Popup/containers/WhitelistSites.tsx
```diff
@@ -10,38 +10,46 @@ import {
...
+  const configQuery = useConfigQuery();
+  const toast = useToast();
+  const configService = useService('configService');
 
-  const removeWhitelistSiteMutation = useMutation({
-    mutationFn: (url: string) => configService.removeWhitelistItem({ url: url }),
+  const removeWhitelistItemMutation = useMutation({
+    mutationFn: (host: string) => {
+      return configService.removeWhitelistItem({ host }) as Promise<void>;
+    },
```
homura, line 31:
```suggestion
    mutationFn: (host: string) => {
      return Promise.resolve(configService.removeWhitelistItem({ host }));
    },
```

homura (approved):
```ts
function useSomeQuery(): [UseQueryResult<...>, InvalidateQuery];
```

using tuple to organize hooks may also be an option

PR #100, opened 2023-03-01
feat: supported event hub

homura:
This PR added an `EventHub` module to support publish/subscribe event between modules and `InjectedCkb`

### In Injected Script

```js
ckb.on('networkChanged', (networkName) => {
  console.log('network changed', networkName)
})

await window.ckb.request({
  method:'debug_updateConfig', 
  params: { selectedNetwork: 'testnet' }
})

// console will output "network changed ckb_testnet"
```

### In Modules

```ts
function myRegisteredModule({ eventHub }) {
  // ...
  eventHub.emit('networkChanged');
}
```

## Deprecated `ckb.enable` and `ckb.isEnabled`

image

`InjectedCkb` should be a pure client, we should not treat it as an SDK, otherwise it may be hard to keep it light and simple, this PR refactored `InjectedCkb` into a pure `EventClient` + `RpcClient`, `InjectedCkb` will expose only three methods

```ts
interface InjectedCkb {
  request(req: { method: string, params?: any }): Promise<any>;
  on(event: string, listener: (...args: any[])): void;
  removeListener(event:string, listener: (...args: any[])): void;
}
```

## Added Debug methods `debug_setConfig` and `debug_getConfig`

Making the debugging easier

### DIFF
--- a/packages/extension-chrome/__tests__/rpc/event.test.ts
+++ b/packages/extension-chrome/__tests__/rpc/event.test.ts
@@ -11,14 +11,15 @@ it('should the networkChanged fired when config has updated', async () => {
 
   ckb.on('networkChanged', mockListener);
   await ckb.request({ method: 'debug_setConfig', params: { selectedNetwork: 'testnet' } });
+  await ckb.request({ method: 'debug_setConfig', params: { selectedNetwork: 'testnet' } });
+  await ckb.request({ method: 'debug_setConfig', params: { selectedNetwork: 'testnet' } });
   await asyncSleep(50);
 
+  expect(mockListener).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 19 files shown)


IronLu233 (approved):

packages/extension-chrome/__tests__/rpc/event.test.ts
```diff
@@ -0,0 +1,29 @@
+import { createTestRpcServer } from './helper';
+import { asyncSleep } from '@ckb-lumos/e2e-test/src/utils';
+import { createInMemoryStorage } from '../../src/services/storage';
+
+it('should the networkChanged fired when config has updated', async () => {
```
zhangyouxin, line 5: Nice simple example demonsting how to use this feature.

PR #102, opened 2023-03-03
feat(extension-chrome): edit networks

IronLu233:


homura (commented):
We should add a limitation to avoid changing the `selectedNetwork`'s URL, since watchtower may still use the old URL

packages/extension-chrome/src/pages/Popup/containers/WhitelistSites.tsx
```diff
@@ -81,7 +81,7 @@ export const WhitelistSites: FC = () => {
         overflowY="auto"
         padding="30px 20px"
         as={WhiteAlphaBox}
-        spacing="16px"
+        spacing="12px"
```
IronLu233, line 84: By the way, fix stack space not correct.

packages/extension-chrome/src/pages/Popup/containers/Network/EditNetwork.tsx
```diff
@@ -0,0 +1,153 @@
...
+        })().catch(() => {});
+      }
+    }
+  }, [mode, id, configService, setValue, trigger]);
+
+  const addNetworkMutation = useMutation({
+    mutationFn: ({ name, rpcUrl }: EditNetworkFormState) => {
+      return configService.addNetwork({
+        network: { displayName: name, rpcUrl, id: nanoid(), networkName: name },
+      }) as Promise<void>;
```
homura, line 85: avoid `as` when possible

packages/extension-chrome/src/pages/Popup/containers/Network/EditNetwork.tsx
```diff
@@ -0,0 +1,153 @@
...
+        (async () => {
+          const networks = await configService.getNetworks();
+          const network = networks.find((network) => network.id === id);
+          if (network) {
+            setValue('name', network.networkName);
+            setValue('rpcUrl', network.rpcUrl);
+            // manually trigger validation
+            await trigger();
+          }
+        })().catch(() => {});
```
homura, line 76: we can tip for error here instead of just catching without doing anything. Or if we do want to leave an empty catch block, a good practice is to leave a comment explaining why an empty catch block is needed

packages/extension-chrome/src/pages/Popup/containers/Network/EditNetwork.tsx
IronLu233: I agree because the return value is PromiseLike, not totally promise.

homura (approved):

PR #105, opened 2023-03-07
chore: update lumos dep

IronLu233:
update lumos to this branch https://github.com/ckb-js/lumos/tree/72830ef3eb85d49b3b77bc37ecbdc58ada2d380c, which solves it can not request http on service worker.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3837,6 +3837,14 @@
         "node": ">=8"
       }
     },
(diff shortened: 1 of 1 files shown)


homura:
we can switch to https://github.com/ckb-js/lumos/commit/72830ef3eb85d49b3b77bc37ecbdc58ada2d380c instead of a branch to avoid error when the service-worker branch is deleted

PR #110, opened 2023-03-09
feat(extension-chrome): validate RPC methods input parameters

### LINKED ISSUE #107: RPC params validation for `wallet_*` methods
Validators for `wallet_*` methods

```diff
- addMethod('wallet_enable', (params, context) => {
+ addMethod('wallet_enable', pipe(
+   validateWalletEnable,
+   (params, context) => {
+ ))
```

IronLu233:
# This PR add parameter validation for RPC invoke.
Now it covers these APIs
-  wallet_fullOwnership_getLiveCells,
- wallet_fullOwnership_getOffChainLocks,
- wallet_fullOwnership_getOnChainLocks,
- wallet_fullOwnership_signData,
- wallet_fullOwnership_signTransaction

closes #107


packages/extension-chrome/src/rpc/schema/primitives.ts
```diff
@@ -0,0 +1,68 @@
...
+});
+
+const ZOutPoint = z.object({
+  txHash: ZHash,
+  index: ZHexNumber,
+});
+
+export const ZDepType = z.union([z.literal('depGroup'), z.literal('code')]);
+
+export const ZCellDep = z.object({
```
zhangyouxin, line 38: These ckb types are not primitive types, we can put them in another file

IronLu233: Do you means create a new file named `blockchain.ts` for defining RPC schema?

packages/extension-chrome/src/services/ownership/fullOwnership.ts
```diff
@@ -99,9 +99,16 @@ export function createFullOwnershipService({
...
-      });
+      let password: string;
+      try {
+        password = (
+          await platformService.requestSignTransaction({
+            tx: transactionSkeletonToObject(txSkeleton),
+          })
+        ).password;
+      } catch (e) {
+        errors.throwError('User has rejected');
```
zhangyouxin, line 110: New feature: `throw NexusCommonErrors.ApproveRejected()`

packages/extension-chrome/src/rpc/schema/primitives.ts
zhangyouxin: Yes, referring to lumos:

- https://github.com/ckb-js/lumos/blob/develop/packages/base/src/blockchain.ts
- https://github.com/ckb-js/lumos/blob/develop/packages/base/src/primitive.ts

packages/extension-chrome/src/rpc/server.ts
```diff
@@ -1,18 +1,16 @@
...
 export const methods: Record<string, (...args: unknown[]) => unknown> = {};
 export const logger = createLogger();
 
-export function addMethod<K extends keyof RpcMethods>(
-  method: K,
-  handler: (param: RpcMethods[K]['params'], context: ServerParams) => RpcMethods[K]['result'],
-): void {
-  Object.assign(methods, { [method]: handler });
+export function addMethod<K extends keyof RpcMethods>(method: K, handler: RPCMethodHandler<K>): void {
+  Object.assign(methods, { [method]: bindSchemaValidator(method, handler) });
```
homura, line 13: It would be better to put validator and RPC method implementation together instead of put all validator together, since when we implement an RPC method, usually we implement its parameter validator along with it 

We can just keep it simple, stupid

```ts

declare function assertSomething(x: unknown): x is Something;

addMethod('some_method', (params) => {
  assertSomething(params)
});
```
At the same time, I realized from this PR that the `addMethod`'s function signature is not "honest", it should actually be 

```ts
function addMethod(
  method: string, 
  (params: unknown, context: NexusContext)
  // 👆
  // the params is from client so it should no be trusted, untyped
): void;
```

packages/extension-chrome/src/rpc/schema/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+type ValidatePayload<S extends AnyRpcSchema, P extends Record<any, any>> = ParameterEqual<SchemaFirstParameter<S>, P>;
+type _cases = [
+  Expect<ValidatePayload<typeof wallet_fullOwnership_getLiveCells, GetLiveCellsPayload>>,
+  Expect<ValidatePayload<typeof wallet_fullOwnership_getOffChainLocks, GetOffChainLocksPayload>>,
+  Expect<ValidatePayload<typeof wallet_fullOwnership_getOnChainLocks, GetOnChainLocksPayload>>,
+  Expect<ValidatePayload<typeof wallet_fullOwnership_signData, SignDataPayload>>,
+
+  Expect<ValidatePayload<typeof wallet_fullOwnership_signTransaction, SignTransactionPayload>>,
+];
```
homura, line 94: We have already defined some types, so avoid writing two sets of code to check the types

packages/extension-chrome/__tests__/rpc/parameterValidate.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    await expect(
+      request('wallet_fullOwnership_signData', {
+        data: '0x1234',
+        lock: { codeHash: '0x114514', hashType: 'data', args: '0x' },
+      }),
+    ).resolves.toBe('signed message');
+  });
+
+  it('wallet_fullOwnership_signTransaction', async () => {
+    const tx = {
```
homura, line 24: while mocking RPC requests can be more complete, it may not be as necessary and may even make testing complicated, as it would require us to mock the entire transaction

we can just add one RPC test to check if the RPC handler checks the parameters, and some unit test to check `HexString`, `HexNumber`, etc.

IronLu233: Finally, I split the integration test. and add a mock module for schema. when calls
```ts
jest.mock('../../src/rpc/schema');
```
All parameter check will be skipped.

packages/extension-chrome/src/rpc/schema/index.ts
```diff
@@ -0,0 +1,84 @@
...
+);
+
+const walletMethodSchemas = {
+  // wallet_enable,
+  wallet_fullOwnership_getLiveCells,
+  wallet_fullOwnership_getOffChainLocks,
+  wallet_fullOwnership_getOnChainLocks,
+  wallet_fullOwnership_signData,
+  wallet_fullOwnership_signTransaction,
+};
```
homura, line 54: it is not a good idea to put all validators together, since contributor may forget to bind the validator to the added method, can you please refactor it like `addMethod`?

```ts
addValidator('some_method', (params) => {

})
addMethod('some_method', () => {
 ...
})
```

IronLu233: image

Does it seems like this? I change the implement but test code are waiting for refactor.

packages/extension-chrome/src/rpc/schema/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const wrapped: RPCMethodHandler<T> = async (param, context) => {
+    const impl = schema.implement(handler);
+
+    try {
+      return await impl(param, context);
+    } catch (e) {
+      if (e instanceof ZodError) {
+        // TODO: format human readable error message of zod error
+        throw NexusError.create({ message: 'Invalid params', data: e });
```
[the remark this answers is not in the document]
IronLu233: https://www.npmjs.com/package/zod-validation-error

I want use this library for formatting error message.

packages/extension-chrome/src/rpc/schema/__tests__/primitive.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(() => schema.parse(value)).toThrow();
+  }
+
+  it('BytesLike', () => {
+    // arraylike buffer
+    expectParseSuccess(ZBytesLike, 'aabbccc');
+    expectParseSuccess(ZBytesLike, new Uint8Array([1, 2, 3]));
+    expectParseSuccess(ZBytesLike, new Uint8Array([1, 2, 3]).buffer);
+
+    expectParseFailure(ZBytesLike, 1);
```
zhangyouxin, line 19: If there is a test case that shows the error data path generated by  `zod-validation-error` will be great

packages/extension-chrome/src/rpc/server.ts
```diff
@@ -1,20 +1,50 @@
...
+export function addMethod<K extends keyof RpcMethods>(method: K, handler: RPCMethodHandler<K>): void {
   Object.assign(methods, { [method]: handler });
 }
 
+/**
+ * add a zod schema to validate the method's arguments
+ * @param method the method name, must be the key of {@link RpcMethods}
+ * @param argSchema the schema of the method's first argument
+ */
+export function addMethodValidator<TKey extends keyof RpcMethods, TArg extends RpcMethods[TKey]['params']>(
```
homura, line 23: it can simplify with

```ts
declare function addMethodValidator<K extends keyof RpcMethods>(
  method: K,
  validator: ZodType<RpcMethods[K]['params']>
)
```

packages/extension-chrome/src/rpc/schema/__mocks__/index.ts
```diff
@@ -0,0 +1,3 @@
+export function bindSchemaValidator(method: unknown, handler: unknown): unknown {
```
homura, line 1: we can rm it if it is unnecessary

packages/extension-chrome/src/rpc/schema/__tests__/index.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+describe.skip('Rpc schema validator', () => {
+  function getWrappedHandler(method: keyof RpcMethods) {
+    return bindSchemaValidator(method, handler) as (...args: any[]) => any;
+  }
+
+  const handler = jest.fn();
+  beforeEach(() => {
+    handler.mockClear();
+  });
```
homura, line 16: it looks like this is a meaningless test, it actually tests the jest itself

```ts
const fn = jest.fn()
fn.mockReturn(someThing)

expect(fn).calledWith(someThing)
```

packages/extension-chrome/src/rpc/server.ts
IronLu233: No, we can't. because if the schema has extra validation field, only `ZodType<RpcMethods[K]['params']>`Can not validate it.

Doing this here is on purpose.

packages/extension-chrome/__tests__/rpc/wallet.test.ts
```diff
@@ -19,9 +19,15 @@ describe('RPC wallet_enable', () => {
     await expect(request('wallet_enable')).resolves.not.toThrowError();
   });
 });
+
 describe('RPC wallet_fullOwnership', () => {
+  let whitelistBypassServer = createTestRpcServer({ skipWhitelist: true });
+  beforeEach(() => {
+    whitelistBypassServer = createTestRpcServer({ skipWhitelist: true });
+  });
```
homura, line 27: why we need to modify these tests?

IronLu233: For making RPC server parameter validation configurable.

packages/extension-chrome/src/rpc/schema/__tests__/index.test.ts
homura, line 1: move them to `__tests__`

packages/extension-chrome/__tests__/rpc/wallet.test.ts
homura: I mean, why does `skipWhitelist` have anything to do with parameter validation

IronLu233: Oh, sorry, I misunderstand. When I run single unit test on this describe suite, it will throw a error.
```
Error: host "nexus-wallet.io" is not in the whitelist, please call "wallet_enable first"Jest
```

This suite depends on previous suite named [`RPC wallet_enable`](https://github.com/ckb-js/nexus/pull/110/files#diff-28ecb711338035a5ae0202e91ed038df015b0f79acb12859a2f771ea13a4dc1dR6).

IronLu233: > I don't think we should modify the previous test case

But how to satisfy the "Run single suite" demand?

packages/extension-chrome/src/rpc/server.ts
homura: you can image that 

```
f(a) = b
a is known
f is known

so b is known

a -> K
f -> RpcMethods[K]['params']
```

packages/extension-chrome/__tests__/rpc/wallet.test.ts
IronLu233: **No, I confirm it is a unit test bug.**

If I skip the first suite, the test will failed.
image

I use the main branch on upstream repo, and just skip `RPC_wallet_enable`

image

packages/extension-chrome/src/rpc/server.ts
IronLu233: I means Typescript's type guard is not `equal`.

For example, my origin schema is
``` ts
{ change: z.enum(['external', 'internal']).optional() }
```
But one day I add a new required field named change2 by mistake.
```ts
{ change: z.enum(['external', 'internal']).optional() }
```
only `ZodType<RpcMethods[K]['params']>` can not check this mistake.

packages/extension-chrome/src/rpc/schema/__tests__/index.test.ts
IronLu233: moved.

IronLu233:
image
Also, error message becomes more human-readable.

packages/extension-chrome/__tests__/rpc/schema/index.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      ],
+      outputsData: ['0x3c484c0000000000', '0x'],
+      headerDeps: ['0xaffb65e2f064eb039c624fc287d40bf60aae42cb7d3e6985cc102260f152deec'],
+      hash: '0x1ea34c7b6735689a9661f1a43ece11a866e47718817d38e0c5b92e5a973df739',
+      version: '0x0',
+      witnesses: [
+        '0x55000000100000005500000055000000410000007037d390c84ad7c9b44d62834e6f0e7f083ee5cedbc12b152392cc7fd00f56aa42ef412d61ba24e09fb19ea952b2996883a861bceffa7cba92df2bab224cc09d00',
+        '0x55000000100000005500000055000000410000000e399817a6120495314b2c13738a84b5e5bdaef8be69b15af924c4ebbaf0373c40ae558f9370d94c72b19ecd697999ae8b73055876d850c3e05ea682664ce67a00',
+      ],
+    } as Transaction;
```
homura, line 99: can we use `const tx: Transaction` instead of `as` here?

packages/extension-chrome/__tests__/rpc/middleware.test.ts
```diff
@@ -45,3 +48,26 @@ describe('Middlewares#errorMiddleware', () => {
     expect(res2?.error?.data).toMatchObject({ method: 'some_method', params: [] });
   });
 });
+
+describe('Middlewares#parameterValidationMiddleware', () => {
+  let server: JSONRPCServer<ServerParams>;
+  beforeEach(() => {
+    server = new JSONRPCServer<ServerParams>();
+    server.applyMiddleware(parameterValidateMiddleware);
```
homura, line 56: a context-free design makes the code more testable

```ts

const server = new JSONRpcServer() 
const server = server.addMethod('validate_str', jest.fn(() => 'validate_str'))
const server = server.addMethod('validate_none', jest.fn(() => 'validate_none'))
server.apply(createParameterValidatorMiddlewire({
  validate_string: z.string()
}))

it('should skip validating if no matching validator', () => {
  await server.receive(...).resolves.toMatchObject({result: 'OK'})
})

it('should validate if matching validator',() => {
  await server.receive(...).resolvce.toMatchObject({ error:... })
})

```

packages/extension-chrome/__tests__/rpc/middleware.test.ts
```diff
@@ -45,3 +48,26 @@ describe('Middlewares#errorMiddleware', () => {
...
+  });
+
+  it('should skip when method have no validator', async () => {
+    server.addMethod('insert_Paimon', jest.fn());
```
homura, line 60: a clear method name makes the test approach easy to understand, e.g. `no_validator`, `str_validator`

packages/extension-chrome/src/rpc/schema/index.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const ZGetPaginateItemsPayload = z.object({
+  cursor: z.string().optional(),
+});
+export const ZFilterPayload = z.object({ change: z.enum(['external', 'internal']).optional() });
+
+export const ZGetOffChainLocksPayload = ZFilterPayload;
+export const ZGetLiveCellsPayload = ZGetPaginateItemsPayload;
+export const ZGetOnChainLocksPayload = ZGetPaginateItemsPayload.merge(ZFilterPayload);
+export const ZSignDataPayload = z.object({ data: ZBytesLike, lock: ZScript });
```
homura, line 13: `data: ZHexString` here is fine, we can keep the same with CKB's RPC

packages/extension-chrome/src/rpc/schema/primitives.ts
```diff
@@ -0,0 +1,21 @@
...
+      'length' in (value as ArrayLike<number>) &&
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      typeof (value as ArrayLike<number>)[Symbol.iterator as any] === 'function'
+    );
+  },
+  { message: 'Invalid ArrayLike<number>' },
+);
+
+const ZArrayBuffer = z.custom<ArrayBuffer>((value) => value instanceof ArrayBuffer, { message: 'Invalid ArrayBuffer' });
+export const ZBytesLike = ZArrayLikeNumber.or(ZArrayBuffer).or(z.string());
```
homura, line 18: So many different types are actually representing `bytes`, but so many expressions may cause confusion, we can only keep `HexString` to describe `bytes` for RPC, which is also consistent with CKB's RCP style

packages/extension-chrome/__tests__/rpc/wallet.test.ts
```diff
@@ -20,6 +20,13 @@ describe('RPC wallet_enable', () => {
   });
 });
 describe('RPC wallet_fullOwnership', () => {
+  beforeAll(async () => {
+    const { request, factory } = createTestRpcServer();
+    const platformService = factory.get('platformService');
+    jest.spyOn(platformService, 'requestGrant').mockImplementation(() => Promise.reject());
+    await request('wallet_enable');
```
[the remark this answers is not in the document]
homura: I got it, the `whitelistMiddleware` was triggered by previous case, so reject in `jest.spyOn(platformService, 'requestGrant').mockImplementation(() => Promise.reject());` will no longer be triggered. But why we need to mock and reject here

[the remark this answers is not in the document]
IronLu233: I copied some code on previous suit and run the test, but when I want to check it, I miss the parameter to limit test suite. It's a mistake, I will update it.

homura (approved):

packages/extension-chrome/src/rpc/schema/primitives.ts
```diff
@@ -0,0 +1,11 @@
+import { z, ZodType } from 'zod';
+
+export const ZHexString = z.string().regex(/^0x([0-9a-fA-F][0-9a-fA-F])*$/, { message: 'Invalid hex string' });
+
+/**
+ * BytesLike type
+ * @description For some serialization reasons, it only accepts hex string
+ */
+export const ZBytesLike: ZodType<z.infer<typeof ZHexString> | ArrayLike<number> | ArrayBuffer> = ZHexString;
```
ast077, line 9: ok, nice type manipulation skill here, since we want to keep alignment with CKB rpc, can we just use `ZHexString` in `signDataPayload` rather than using comments here

zhangyouxin, line 9: ok, nice type manipulation skill here, since we want to keep alignment with CKB rpc, can we just use ZHexString in signDataPayload rather than using comments here

packages/extension-chrome/src/rpc/server.ts
```diff
@@ -1,20 +1,44 @@
...
+ * add a zod schema to validate the method's arguments
+ * @param method the method name, must be the key of {@link RpcMethods}
+ * @param argSchema the schema of the method's first argument.When it's type is `never`, it means the schema is not same as `RpcMethods[TKey]['params']`
+ */
+export function addMethodValidator<TKey extends keyof RpcMethods, TArg extends ZodType>(
+  method: TKey,
+  argSchema: ObjectEquals<RpcMethods[TKey]['params'], z.infer<TArg>> extends true ? TArg : never,
+): void {
+  if (!methods[method as string]) {
+    errors.throwError(`Method ${method} is not registered yet. Please call \`addMethod\` first.`);
```
zhangyouxin, line 31: Looks like a common error to me, is it good to add a `NexusCommonErrors.UnknownRpcMethod` common error?

packages/extension-chrome/src/rpc/schema/primitives.ts
IronLu233: We can't, if we do these, the type check of (`addMethodValidator('wallet_fullOwnership_signData', ZSignDataPayload)`)[https://github.com/ckb-js/nexus/pull/110/files#diff-a09671bdb92e37d0c3072838a10894c1ac901cc55a0ddb5dff74c96dba69600fR59] will failed. Here, the comment tells other developer, it is on purpose. The type is `HexString | ArrayLikeNumber<number> | ArrayBuffer`, but the implements only accept hex string.

I will update the comment, because in potential, it will make other develop confused.

packages/extension-chrome/src/rpc/server.ts
```diff
@@ -1,20 +1,44 @@
...
 
+type ObjectEquals<X, Y> = X extends Y ? (Y extends X ? true : false) : false;
+/**
+ * add a zod schema to validate the method's arguments
+ * @param method the method name, must be the key of {@link RpcMethods}
+ * @param argSchema the schema of the method's first argument.When it's type is `never`, it means the schema is not same as `RpcMethods[TKey]['params']`
+ */
+export function addMethodValidator<TKey extends keyof RpcMethods, TArg extends ZodType>(
+  method: TKey,
+  argSchema: ObjectEquals<RpcMethods[TKey]['params'], z.infer<TArg>> extends true ? TArg : never,
```
zhangyouxin, line 28: Nice type manipulation skill

packages/extension-chrome/src/rpc/server.ts
IronLu233: That is a good idea.

packages/extension-chrome/src/rpc/schema/primitives.ts
zhangyouxin: ok, sounds reasonable

packages/extension-chrome/src/rpc/server.ts
IronLu233: But I think this error is not caused by user, use `logger.error` is enough.
```ts
  if (!methods[method as string]) {
    logger.error(
      `Method ${method} is not registered yet. Please call \`addMethod\` first. This addMethodValidator call will be ignored.`,
    );
    return;
  }

```

zhangyouxin: we can refer to ckb node rpc error: image

IronLu233: I means it is a bug caused by developer, not caused by Nexus user, `NexusCommonErrors` is to tell user what's wrong on their inputs. So we shouldn't use it.

zhangyouxin (approved):

PR #111, opened 2023-03-09
feat: make rpc error more friendly

homura:
This PR make the JSON RPC error response more friendly

```ts
interface Error<T> {
  // for both end-user and developer to know what 
  message: string,
  // for developer, more details about the error, 
  // e.g. when a cell cannot be resolved, the `data` will be a outPoint
  data?: T
}

```

## How 

```js
if (incorrectCondition) {
  throw NexusError.create({ message: 'Error message' })
}

// or
if (anotherCondition) {
  throw NexusCommonErrors.RequestTimeOut({
    method: 'get_transactions', 
    params: [...]
  });
}
```


homura:
it is unnecessary to modify backend.ts in this PR, we can migrate to https://github.com/ckb-js/nexus/pull/112

PR #112, opened 2023-03-10
feat: make rpc error more friendly

homura:
This PR make the JSON RPC error response more friendly

```ts
interface Error<T> {
  // for both end-user and developer to know what 
  message: string,
  // for developer, more details about the error, 
  // e.g. when a cell cannot be resolved, the `data` will be a outPoint
  data?: T
}
```

checkout the [error handling](https://github.com/ckb-js/nexus/blob/friendly-rpc-error-new/docs/error.md) for more detail

## How 

```js
if (incorrectCondition) {
  throw NexusError.create({ message: 'Error message' })
}

// or
if (anotherCondition) {
  throw NexusCommonErrors.RequestTimeOut({
    method: 'get_transactions', 
    params: [...]
  });
}
```

### DIFF
--- a/packages/extension-chrome/src/errors/NexusError.ts
+++ b/packages/extension-chrome/src/errors/NexusError.ts
@@ -1,0 +1,27 @@
+/**
+ * an error to make RPC could return a friendly error
+ */
+export class NexusError<T = unknown> extends Error {
+  __NEXUS_ERROR__ = true;
+  code: number;
+  data?: unknown;
+
+  private constructor({ code, message, data }: { code: number; message: string; data?: T }) {
+    super(message);
+    this.code = code;
+    this.data = data;
+  }
+
+  static create<T>(opt: { code?: number; message: string; data?: T } | string): NexusError<T> {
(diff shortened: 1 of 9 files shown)


packages/extension-chrome/src/errors/NexusError.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * an error to make RPC could return a friendly error
+ */
+export class NexusError<T = unknown> extends Error {
```
IronLu233, line 4: Here, I have one question:

If I want post an error when RPC parameter Error, It means that I should extend this generic class? Or use `create` method?

homura: Just append 

```diff
const NexusErrors = {

+ ParamValidationError: (err: {
+   message: ErrorMessage, 
+   data: ZodValidationObject 
+ }) => NexusError<ZodValidationObject>
```

IronLu233 (approved):

zhangyouxin (approved):
LGTM

PR #122, opened 2023-03-13
feat(extension-chrome): Use Chrome extension favicon API for whitelist and grant

IronLu233:


homura:
it is ready for review?

homura:
BTW, can you help to remove favicons from the `TrustedHost`

zhangyouxin:
It would be great if some reference links are present in PR description so that I can understand how this PR works.

packages/extension-chrome/src/pages/Components/SiteFavicon.tsx
```diff
@@ -0,0 +1,15 @@
+import { Image, ImageProps } from '@chakra-ui/react';
+import React, { FC, useMemo } from 'react';
+
+export const SiteFavicon: FC<
+  Omit<ImageProps, 'src' | 'w' | 'h' | 'width' | 'height'> & { host: string; size?: number }
+> = ({ host, size = 32, ...rest }) => {
+  const faviconURL = useMemo(() => {
+    const url = new URL(chrome.runtime.getURL('/_favicon/'));
```
homura, line 8: we should avoid using the `chrome` API directly, it is better to add a new method to `PlatformService` to support favicon fetching

```ts
declare function getFavicon(options: { host: string, size?: string }): Promise<string>
````

**Review by homura — APPROVED** [review]

File: packages/extension-chrome/__tests__/helpers/mockPlatformService.ts [hunk 2]
```diff
@@ -5,11 +5,11 @@ export const MOCK_PLATFORM_PASSWORD = '12345678';
 
 // TODO: move it to @nexus-wallet/testkit
 export const mockPlatformService: PlatformService = /* istanbul ignore next */ {
-  /* istanbul ignore next */
-  getActiveSiteInfo: async () => undefined,
+  /* istanbul ignore next */ getActiveSiteInfo: async () => undefined,
```
homura, line 8:
```suggestion
   getActiveSiteInfo: /* istanbul ignore next */ async () => undefined,
```

PR #127, opened 2023-03-14
feat: ownership provider to work with Lumos

homura:
- [x] `injectCapacity`
- [x] `payFee`
- [x] `collector`
- [x] `signTransaction`

update(2023-03-20):

- we should change constructor from `{ckb: InjectedCkb}` to `{ckb: InjectedCkb , rpcUrl: string }` since the `ScriptConfig` is required for the `Provider` https://github.com/ckb-js/nexus/blob/abadb896648b0587254b662535dd26d764f02540/packages/extension-chrome/src/services/ownership/backend/backendUtils.ts#L66-L87

[nexus--canary.127.4635976068.zip](https://github.com/ckb-js/nexus/releases/download/v0.0.0-canary/nexus--canary.127.4635976068.zip)


packages/ownership-providers/examples/transfer.ts
```diff
@@ -0,0 +1,58 @@
+import { FullOwnershipProvider } from '../src';
+import { TransactionSkeleton } from '@ckb-lumos/helpers';
+import { createCell } from '@ckb-lumos/experiment-tx-assembler';
+import { Cell, Script } from '@ckb-lumos/base';
+import { BIish } from '@ckb-lumos/bi';
+import produce from 'immer';
+
+declare const recipient: Script;
+declare const provider: FullOwnershipProvider;
```
IronLu233, line 9: How to get the provider instance? it seems to be:
```ts
const provider = new FullOwnershipProvider(globalThis.ckb)
```

homura:
```ts
new FullOwnershipProvider({ ckb: globalThis.ckb })
``` 

we can check out from [FullOwnershipProvider.ts](https://github.com/ckb-js/nexus/blob/91fdd353a0c84866f5b5c9865cea49db2726cef9/packages/ownership-providers/src/FullOwnershipProvider.ts#L33)

packages/ownership-providers/examples/transfer.ts
```diff
@@ -0,0 +1,58 @@
+import { FullOwnershipProvider } from '../src';
+import { TransactionSkeleton } from '@ckb-lumos/helpers';
+import { createCell } from '@ckb-lumos/experiment-tx-assembler';
+import { Cell, Script } from '@ckb-lumos/base';
+import { BIish } from '@ckb-lumos/bi';
+import produce from 'immer';
+
+declare const recipient: Script;
```
IronLu233, line 8: An idea: a test case can be based on this example.

IronLu233:
I want to reuse lumos prepare signing entries method.
https://github.com/ckb-js/lumos/blob/e73ca0502b0feb3e98be848f76325a912052d0f0/packages/common-scripts/src/helper.ts#L186-L279

Thus, my code will seems like
```ts
  /**
   * request wallet to sign a transaction skeleton
   * @param txSkeleton The transaction skeleton, you can create it from transaction object via `@ckb-lumos` {@link createTransactionFromSkeleton}
   * @returns The signed transaction skeleton. To get the signed transaction object, please use {@link sealTransaction} with empty sealingContents(`[ ]`).
   */
  async signTransaction(txSkeleton: TransactionSkeletonType): Promise<TransactionSkeletonType> {
    const config = getConfigFromRpcUrl();

    prepareSigningEntries(txSkeleton, config);
    const groupedSignature = await this.ckb.request({
      method: 'wallet_fullOwnership_signTransaction',
      params: { tx: createTransactionFromSkeleton(txSkeleton) },
    });

    txSkeleton = txSkeleton.update('witnesses', (witnesses) => {
      return witnesses.map((witness, index) => {
        const [, signature] = groupedSignature[index];
        if (!signature) return witness;
        const witnessArgs = WitnessArgs.unpack(witness);
        return bytes.hexify(
          WitnessArgs.pack({
            ...witnessArgs,
            lock: signature,
          }),
        );
      });
    });

    txSkeleton = txSkeleton.update('signingEntries', (signingEntries) =>
      signingEntries.splice(0, groupedSignature.length),
    );

    return txSkeleton;
  }

```

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  /**
+   * Pay the transaction fee
+   * @param txSkeleton
+   * @param options
+   */
+  async payFee({
+    txSkeleton,
+    options = { autoInject: true },
```
homura, line 161: If `payers` appears in the `options`, the `autoInject: true` should be passed in explicitly

Because if the developer thinks that the fee will ONLY be paid by `payers`, and `autoInject` is not found during the test, it may cause an unexpected problem. The worst-case is the transaction is assembled and failed when he `autoInject: false` is defaults when `payers` is set.

Additionally, we need to add more clear documentation for this function

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,273 @@
...
+      if (!injected) {
+        errors.throwError(autoInject ? 'No cell sufficient to pay fee' : 'No payer available to pay fee');
+      }
+      currentTransactionSize = getTransactionSizeByTx(createTransactionFromSkeleton(txSkeletonWithFee));
+    }
+
+    return txSkeletonWithFee;
+  }
+
+  async *collector({ lock }: { lock?: Script } = {}): AsyncIterable<Cell> {
```
zhangyouxin, line 208: pro usage, not easy to understand though

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    return txSkeleton;
+  }
+
+  /**
+   * Pay the transaction fee
+   * @param txSkeleton
+   * @param options
+   */
+  async payFee({
```
zhangyouxin, line 159: sometimes tx fee can be charged by deducing capacity in some output cell, we don't need to support that option for now?

zhangyouxin: One scenario is that, Alice has got 10_000 CKB from faucet, when Alice wants to transfer some CKB to Bob, he just can't find another cell to pay fee

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,273 @@
...
+   */
+  async signTransaction(txSkeleton: TransactionSkeletonType): Promise<TransactionSkeletonType> {
+    const config = await this.getLumosConfig();
+    prepareSigningEntries(txSkeleton, { config });
+    const groupedSignature = await this.ckb.request({
+      method: 'wallet_fullOwnership_signTransaction',
+      params: { tx: createTransactionFromSkeleton(txSkeleton) },
+    });
+
+    txSkeleton = txSkeleton.update('witnesses', (witnesses) => {
```
zhangyouxin, line 238: This works when all input cells are controlled by wallet.
But I would like to leave a note here that if rule-based-ownership is interated, that means, when some input cells are not controlled by wallet, we should compare returned `lock` in groupedSignature and actual lock in the tx of corresponding index. 

| inputs    | lock                   | where to put signature |
|-----------|------------------------|------------------------|
| inputs[0] | acp-lock               | "0x"                   |
| inputs[1] | wallet-controlled-lock | real-signature         |

packages/ownership-providers/src/FullOwnershipProvider.ts
IronLu233: Do you means `autoInject` never can never become `false`?

packages/ownership-providers/src/FullOwnershipProvider.ts
IronLu233: It will take other problems. Although I consider it but finally, I make it's implementation same as lumos.

packages/ownership-providers/src/FullOwnershipProvider.ts
homura: I mean the default options would be 

```ts
type DefaultOptions =
  | { autoInject: true }
  // if `payers` are defined, the `autoInject: true` should passed explicitly 
  | { payers: LockLike[]; autoInject: false }
```

IronLu233: With function overload, It can be more clear.
```ts
  /**
   * Pay the transaction fee using the wallet owned lock
   * @param params.txSkeleton The transaction skeleton
   * @param params.options.feeRate the fee rate, if omitted the fee rate will be calculated automatically
   * @example
   * ```ts
   * const provider = new FullOwnershipProvider({ ckb });
   * // auto calculate fee rate and use wallet owned lock to pay fee
   * const txSkeleton = await provider.payFee({ txSkeleton });
   * ```
   */
  payFee(params: {
    txSkeleton: TransactionSkeletonType;
    options?: { feeRate?: number } & PayByAuto;
  }): Promise<TransactionSkeletonType>;

  /**
   * Pay the transaction fee using the specified lock
   * @description `options.payers` and options `options.autoInject` are required. If `options.autoInject` is true, wallet owned lock can be payer as fallback
   * @param params.txSkeleton The transaction skeleton
   * @param params.options.payers Only cell with these lock scripts can be used to pay fee
   * @param params.options.autoInject if true, wallet owned lock can be payer as fallback
   * @example
   * ```ts
   * const provider = new FullOwnershipProvider({ ckb });
   * // auto inject capacity when payers can not cover the fee
   * const txSkeleton = await provider.payFee({ txSkeleton, payers: [payer1, payer2], autoInject: true });
   * // throw error when payers can not cover the fee
   * const txSkeleton = await provider.payFee({ txSkeleton, payers: [payer1, payer2], autoInject: false });
   * ```
   */
  payFee(params: {
    txSkeleton: TransactionSkeletonType;
    options: { feeRate?: number } & PayByPayers;
  }): Promise<TransactionSkeletonType>;

```
The result on my IDE: 
image

homura: We should give up the default `autoInject` for users, this default behavior brings more confusion than convenience, a good API design should be able to go beyond the documentation

homura: So the options could be

```ts
type Options = {
  autoInject: boolean;
  payers?: LockScriptLike[];
  feeRate?: BIish;
}
```

This also makes the API clearer and less need for redundant explanations

homura:
Insufficient coverage of unit tests, can you add tests?

packages/ownership-providers/examples/transfer.ts
```diff
@@ -0,0 +1,58 @@
...
+async function transferCkb(): Promise<void> {
+  // create an empty transaction skeleton
+  let txSkeleton = TransactionSkeleton();
+
+  // define the recipient cell
+  const recipientCell = createCell({ capacity, lock: recipient });
+  txSkeleton = txSkeleton.update('outputs', (outputs) => outputs.push(recipientCell));
+
+  txSkeleton = await provider.injectCapacity(txSkeleton, { amount: capacity });
+  txSkeleton = await provider.payFee(txSkeleton);
```
zhangyouxin, line 21: parameter doesn't match, need to pass in options here

IronLu233: amended

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,306 @@
...
+  /**
+   * if `true`, the fee can be paid by the other wallet owned cells. if {@link PayFeeOptions#payers} is specified, payers lock have higher priority.
+   */
+  autoInject: boolean;
+};
+export type PayBy = PayByPayers | PayByAuto;
+/** Pay by the specified payers */
+export type PayByPayers = { payers: LockScriptLike[]; autoInject: boolean };
+/** Pay by inject automatically */
+export type PayByAuto = { autoInject: true };
```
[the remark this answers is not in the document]
IronLu233: removed

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,302 @@
...
+  /**
+   * Pay the transaction fee using the specified lock
+   * @param txSkeleton The transaction skeleton
+   * @param options.payers if provided, candidates for paying fee will be filtered by these lock scripts
+   * @param options.autoInject if true, wallet owned lock can be payer as fallback
+   * @param options.feeRate The fee rate, in Shannons per byte. If not specified, the fee rate will be calculated automatically.
+   * @example
+   * ```ts
+   * const provider = new FullOwnershipProvider({ ckb });
+   * const provider = new FullOwnershipProvider({ ckb });
```
zhangyouxin, line 171: duplicated

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,302 @@
...
+          yield cell;
+        }
+      }
+    }
+  }
+
+  /**
+   * request wallet to sign a transaction skeleton
+   * @param txSkeleton The transaction skeleton, you can create it from transaction object via {@link @ckb-lumos/helpers#TransactionSkeleton}
+   * @returns The signed transaction skeleton. Signing message is in `witnesses` field.
```
zhangyouxin, line 259: "The signed transaction skeleton. The signatures are serialized and put into the `witnesses` field."

packages/ownership-providers/src/__tests__/FullOwnershipProvider.test.ts
```diff
@@ -0,0 +1,413 @@
...
+
+    await expect(provider.getOnChainLocks({ cursor: '0' })).resolves.toEqual([]);
+    expect(ckbRequest).toBeCalledWith({
+      method: 'wallet_fullOwnership_getOnChainLocks',
+      params: {
+        cursor: '0',
+      },
+    });
+  });
+});
```
zhangyouxin, line 413: Test cases have sufficient coverage

homura:
can you try the coverage detection method mentioned in this [post](https://community.codecov.com/t/wrong-coverage-percentage/1775/7)?

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,305 @@
+import { BIish, BI } from '@ckb-lumos/bi';
```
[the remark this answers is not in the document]
IronLu233: But this file only export a class named `FullOwnershipProvider`

zhangyouxin (approved):

IronLu233:
> can you try the coverage detection method mentioned in this [post](https://community.codecov.com/t/wrong-coverage-percentage/1775/7)?
```xml
      <line num="138" count="10" type="cond" truecount="1" falsecount="0"/>
```

But it seems the coverage is wrong. jest tells me the `if` statement is full covered.

image

packages/ownership-providers/src/__tests__/FullOwnershipProvider.test.ts
```diff
@@ -0,0 +1,419 @@
...
+
+    function initProviderWithCells(cells: Cell[], offChainLock = offChainLock1) {
+      const provider = new FullOwnershipProvider({} as any);
+      provider.collector = jest.fn().mockImplementation(async function* mock() {
+        for (const cell of cells) {
+          yield cell;
+        }
+      });
+      provider.getOffChainLocks = jest.fn().mockResolvedValue([offChainLock]);
+      return provider;
```
homura: maybe we can use a custom `injectedCkb` instead of mock `FullOwnershipProvider` itself, because

1. the `ckb` object is reusable and testable for another modules(dApp)
2. abstraction is necessary in order to make the software reducing it complexity and efforts, I find that what you are mocking is similarly implemented in the testcase of ownership's

so we should try to abstract a lib for testing [purpose](https://github.com/ckb-js/nexus/pull/161)

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,305 @@
...
+      data: '0x',
+    };
+    const minimalChangeCapacity = minimalCellCapacityCompatible(changeCell);
+
+    let neededCapacity = BI.from(config.amount).add(minimalChangeCapacity);
+    const inputCells: Cell[] = [];
+    const payerLock = config.lock ? await this.parseLockScriptLike(config.lock) : undefined;
+
+    for await (const cell of this.collector({ lock: payerLock })) {
+      inputCells.push(cell);
```
homura, line 136: We should skip using the cell if it is already in `inputs`

```ts
if (inputCells.find(item => 
  bytes.equal(OutPoint.pack(item.outPoint), OutPoint.pack(cell.outPoint)))
) continue
```

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,305 @@
...
+export type PayFeeOptions = {
+  /**
+   * The fee rate, in Shannons per byte. If not specified, the fee rate will be calculated automatically.
+   */
+  feeRate?: BIish;
+
+  /**
+   * Only cell with these lock scripts can be used to pay fee
+   */
+  payers?: LockScriptLike[];
```
homura, line 36: The `payers` is still too ambiguous, and we should probably use a more clear naming such as `byOutPutIndexes` to indicate which cells in the outputs are used to pay the tx fee

```ts
byOutPutIndexes: number[]
```

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,305 @@
...
+   * const txSkeleton = await provider.payFee({ txSkeleton, payers: [payer1, payer2], autoInject: true });
+   * // throw error when payers can not cover the fee
+   * const txSkeleton = await provider.payFee({ txSkeleton, payers: [payer1, payer2], autoInject: false });
+   * ```
+   */
+  async payFee(txSkeleton: TransactionSkeletonType, options: PayFeeOptions): Promise<TransactionSkeletonType> {
+    if ('payers' in options && options.payers?.length === 0 && !options.autoInject) {
+      errors.throwError('no payer is provided, but autoInject is `false`');
+    }
+
```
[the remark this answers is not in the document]
zhangyouxin: The logic of detecting if tx-fee is already paid: https://github.com/ckb-js/nexus/pull/127/files#diff-4212a8e65e25ac3827c0789664613414267d82d12237316f5b3adf09ac533476R209-R215

Is it better to run this logic before validating input parameters?

packages/ownership-providers/src/FullOwnershipProvider.ts
zhangyouxin: skipped: https://github.com/ckb-js/nexus/pull/127/files#diff-4212a8e65e25ac3827c0789664613414267d82d12237316f5b3adf09ac533476R145-R152

packages/ownership-providers/src/__tests__/FullOwnershipProvider.test.ts
zhangyouxin: mocked `ckb`: https://github.com/ckb-js/nexus/pull/127/files#diff-20b325f58e61e13bb6156bd2fb02477d6aba804ad9d079926a9ad2a05a1449b4R74

packages/ownership-providers/src/FullOwnershipProvider.ts
zhangyouxin: updated

packages/ownership-providers/src/FullOwnershipProvider.ts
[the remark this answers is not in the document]
IronLu233: I think validating parameter first is better.

Thinking the user give a invalid parameter, but because of detecting tx fee already paid, it is not throw.

It is not Intuitive.

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+      const currentCell = txSkeleton.get('outputs').get(byOutPutIndex);
+      if (!currentCell) {
+        errors.throwError('`byOutPutIndex` is out of range');
+      }
+      const cellCapacity = BI.from(currentCell.cellOutput.capacity);
+      const minimalCapacity = minimalCellCapacityCompatible(currentCell);
+
+      // How many capacity is needed?
+      // cellCapacity - originFee >= minimalCapacity ? originFee : cellCapacity - minimalCapacity
+      const neededCapacity = cellCapacity.sub(remainFee).gte(minimalCapacity)
```
homura, line 232: `affordCapacity` is more clear here. And if we replace the `remainFee` with the `neededCapacity` may make the code more clear

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+   * ```
+   */
+  async payFee(txSkeleton: TransactionSkeletonType, options: PayFeeOptions): Promise<TransactionSkeletonType> {
+    if (options.byOutPutIndexes?.length === 0 && !options.autoInject) {
+      errors.throwError('no byOutPutIndexes is provided, but autoInject is `false`');
+    }
+    const autoInject = !!options.autoInject;
+    const feeRate = BI.from(options.feeRate || 1000);
+    let currentTransactionSize = getTransactionSizeByTx(createTransactionFromSkeleton(txSkeleton));
+    let txSkeletonWithFee = txSkeleton;
```
homura, line 204: Maybe it is unnecessary since the `txSkeleton` is immutable, `txSkeleton` is fine

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+    const paidFee = calculateFeeCompatible(currentTransactionSize, feeRate).sub(remainFee);
+
+    while (currentTransactionSize > size) {
+      size = currentTransactionSize;
+      const fee = calculateFeeCompatible(size, feeRate).sub(paidFee);
+
+      try {
+        txSkeletonWithFee = await this.injectCapacity(txSkeleton, {
+          amount: fee,
+        });
```
homura, line 262: Every `injectCapacity` will change the skeleton, so the fee will be changed after every `injectCapacity`, it means that we should re-calc fee every time after `injectCapacity`, so we can just try to inject cells to make it possible to cover at least the current fee until enough is paid

```ts
let frozenOutputs = txSkeleton.outputs;
if (neededFee > 0 && autoInject) { 
  txSkeleton = await this.injectCapacity(txSkeleton, { amount: neededFee })
}

return this.payFee(txSkeleton, {
  autoInject: true,
  // just pseudocode, requreid to implement the diff
  byOutputIndexes: txSkeleton.outpus.diff(frozenOutputs)
})
```

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,352 @@
...
+export type PayFeeOptions = {
+  /**
+   * The fee rate, in Shannons per byte. If not specified, the fee rate will be calculated automatically.
+   */
+  feeRate?: BIish;
+
+  /**
+   * Only in these output indexes, can pay fee
+   */
+  byOutPutIndexes?: number[];
```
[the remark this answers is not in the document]
IronLu233: lock removed.

and fixed typo.

packages/ownership-providers/src/FullOwnershipProvider.ts
IronLu233: removed.

packages/ownership-providers/src/FullOwnershipProvider.ts
IronLu233: updated.

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,326 @@
...
+    if (options.byOutputIndexes?.length === 0 && !options.autoInject) {
+      errors.throwError('no byOutputIndexes is provided, but autoInject is `false`');
+    }
+    const autoInject = !!options.autoInject;
+    const feeRate = BI.from(options.feeRate || 1000);
+    let currentTransactionSize = getTransactionSizeByTx(createTransactionFromSkeleton(txSkeleton));
+
+    // feeFromTransactionSize -  (sum(inputsCapacity) - sum(outputsCapacity))
+    let requireFee = calculateFeeCompatible(currentTransactionSize, feeRate).sub(
+      sumCapacity(txSkeleton.get('inputs')).sub(sumCapacity(txSkeleton.get('outputs'))),
```
homura, line 196:
```ts
// FIXME detect if it is a DAO unlock transaction first
```

packages/ownership-providers/src/FullOwnershipProvider.ts
```diff
@@ -0,0 +1,326 @@
...
+
+    for await (const cell of this.collector()) {
+      if (
+        inputCells.find(
+          (item) =>
+            !!item.outPoint && cell.outPoint && bytes.equal(OutPoint.pack(item.outPoint), OutPoint.pack(cell.outPoint)),
+        )
+      ) {
+        continue;
+      }
```
homura, line 141:
```ts
// a cell with lock-only will be injected
if (cell.type || cell.data) {
  continue
}

```

homura:
@IronLu233 could you comment for the `payFee` does not support unlocking DAO transaction yet

PR #128, opened 2023-03-15
feat: separated the protocol from the types

zhangyouxin:


packages/extension-chrome/src/services/ownership/fullOwnership.ts
```diff
@@ -142,8 +143,10 @@ export function createFullOwnershipService({
         'Cannot find script info associated with lock %s, this error is unlikely to occur, have you changed the data in storage or have you manually built the data in storage?',
         payload.lock,
       );
+
+      const prefixedData = `${SIGN_DATA_MAGIC}${bytes.hexify(payload.data).slice(2)}`;
```
homura, line 147: `bytes.concat` is fine

packages/extension-chrome/src/rpc/walletImpl.ts
```diff
@@ -14,7 +16,7 @@ addMethod('wallet_enable', async (_, { getRequesterAppInfo, resolveService }) =>
   const { whitelist } = await configService.getConfig();
 
   const isTrusted = whitelist.find((item) => item.host === host);
-  if (isTrusted) return;
+  if (isTrusted) return response;
```
homura, line 19: can you help to add a test case for this change?

homura:
The re-export in `types` pkg is a bit messy, which will cause the document generation and other functions to not work properly, I will adjust the definition of `types` and `protocols`, this PR can first focus on implementing the protocol adjustment

packages/extension-chrome/src/rpc/walletImpl.ts
zhangyouxin: added: https://github.com/ckb-js/nexus/pull/128/commits/1a088804b1316e7689a3f0e5a31416af73668bd6

homura:
I have implemented the two TODOs that were left in the protocol, so this PR only requires

1. only implement `signData` with magic bytes
2. change the PR title

zhangyouxin:
> I have implemented the two TODOs that were left in the protocol, so this PR only requires
> 
> 1. only implement `signData` with magic bytes
> 2. change the PR title

Closed, move to #139

PR #138, opened 2023-03-22
refactor: disable unsafe any access

### LINKED ISSUE #134: refactor code with no-unsafe-member-access
To reduce the risk of null pointer exceptions in runtime

avoid `any.someProperty`

IronLu233:
# What Changed

Resolve #134 #138

add these ESlint rules. and disable some rules on unit test file.

- '@typescript-eslint/no-implicit-any-catch
- '@typescript-eslint/no-unsafe-argument
- '@typescript-eslint/no-unsafe-assignment'
- '@typescript-eslint/no-unsafe-call'
- '@typescript-eslint/no-unsafe-member-access'
-  '@typescript-eslint/no-unsafe-return'

## Motivation
In our system we have some unsafe access, because they are from 3rd party library. In this PR, unless the developer use `as Type` first, they can not access the prop in the object.

And for error catching, the developer must determine the error type, than can do the sequence thing.
## Change Type

- [ ] `documentation`
- [x] `internal`
- [ ] `patch`
- [ ] `minor`
- [ ] `major`


ecyktdduqli5069 (approved):
LGTM

packages/extension-chrome/src/services/ownership/watchtower.ts
```diff
@@ -185,7 +185,7 @@ export function createWatchtower({
       }
       try {
         await scanAndUpdate();
-      } catch (e) {
+      } catch (e: unknown) {
```
homura, line 188: unnecessary, the catched object defaults to unknown in typescript

packages/extension-chrome/src/services/ownership/backend/backendUtils.ts
```diff
@@ -133,7 +133,7 @@ function createRpcClient(url: string, options?: RpcClientOptions): RpcClient {
     if (res.error !== undefined) {
       throw NexusCommonErrors.RequestCkbFailed(res);
     }
-    return res.result;
+    return res.result as Result;
```
homura, line 136: Unnecessary, this type labeling is just a lie to yourself

IronLu233: In this return statement, if we return `res.result` directly, it will return a "unsafe type".

https://typescript-eslint.io/rules/no-unsafe-return/

IronLu233: updated

packages/extension-chrome/src/services/storage/memory.ts
```diff
@@ -11,11 +11,11 @@ export function createInMemoryStorage<S>(): InMemoryStorage<S> {
   const store = new Map();
 
   return {
-    getItem(key) {
-      const value = store.get(key);
-      if (!value) return value;
+    getItem<K extends keyof S>(key: K) {
```
homura, line 14: we should repeat interface here? `InMemoryStorage<S>` has defined the `getItem`

IronLu233: updated. It is for `as S[K]` but we don't need anymore.

packages/extension-chrome/src/services/ownership/backend/backendUtils.ts
```diff
@@ -155,7 +155,7 @@ function createRpcClient(url: string, options?: RpcClientOptions): RpcClient {
       if (res.error !== undefined) {
         throw NexusCommonErrors.RequestCkbFailed(res);
       }
-      return res.result;
+      return res.result as Result;
```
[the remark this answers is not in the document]
IronLu233: removed

homura (approved):

PR #145, opened 2023-03-23
docs: run a dev CKB blockchain

zhangyouxin:
# What Changed

## Motivation

By default, Nexus wallet is connected to the CKB testnet blockchain, some DApp developers need to test their products quickly, so it will make things easier if they could run a dev CKB blockchain on the local machine, this PR provides a guide. 

preview: https://github.com/ckb-js/nexus/blob/40b5f8b/docs/run-dev-ckb-blockchain.md#how-to-reset-the-ckb-dev-blockchain

## Change Type

- [x] `documentation`
- [ ] `patch`
- [ ] `minor`
- [ ] `major`


homura:
A nice explanation of how to start a dev chain locally

Perhaps we could document an example of how it works with Nexus, for example, adding a derived lock in genesis block and adding a dev chain configuration for Nexus networks

docs/run-dev-ckb-blockchain.md
```diff
@@ -0,0 +1,193 @@
...
++     rpcUrl: 'http://localhost:8114',
++ }
+```
+
+and change default network to `devnet` in [36th line of src/services/internal.ts](https://github.com/ckb-js/nexus/blob/ba5923525ac7706e5e14834b9c68b4652f4a83db/packages/extension-chrome/src/services/internal.ts#L36)
+
+```diff
+- selectedNetwork: 'testnet',
++ selectedNetwork: 'devnet',
+```
```
homura, line 164: I think we have supported now

homura (approved):

PR #157, opened 2023-03-27
fix: connect status is always connected

### LINKED ISSUE #149: connect status is always Connected
image

IronLu233:
# What Changed
closes #149 
## Motivation
In previous version, the connect status indicator can not react when tab changes.
This PR fix it.

image
image

## Change Type

- [ ] `documentation`
- [x] `patch`
- [ ] `minor`
- [ ] `major`


packages/extension-chrome/src/pages/Popup/containers/Home.tsx
```diff
@@ -16,9 +18,48 @@ const FeedbackButton: FC<ButtonProps> = (props) => {
   );
 };
 
+const useConnectedStatus = () => {
+  const configService = useService('configService');
+  const [activeTab, setActiveTab] = useState<chrome.tabs.Tab>();
+
+  const updateCurrentTab = useCallback(async () => {
+    const tabs = await chrome.tabs.query({ active: true, currentWindow: true });
```
homura, line 26: We should avoid using `chrome` API directly since it is hard for testing because we may later unit test the UI as well, or port it to mobile. Can you please replace with `GrantService.isGranted`

Additionally, it is unnecessary to listen to the changes of connect status because when a dApp want to connect to Nexus, a notification page will replace the popup page and the connect status on popup page will be refreshed when it is opened again

packages/extension-chrome/src/pages/Popup/containers/Home.tsx
```diff
@@ -16,9 +18,48 @@ const FeedbackButton: FC<ButtonProps> = (props) => {
   );
 };
 
+const useConnectedStatus = () => {
+  const configService = useService('configService');
+  const [activeTab, setActiveTab] = useState<chrome.tabs.Tab>();
+
+  const updateCurrentTab = useCallback(async () => {
```
IronLu233, line 25: However, seems we can not access grant service via `makeBrowserExtensionModulesFactory`, because `GrantService` is not in `Modules`
```ts
export interface Modules<S = unknown, P = unknown> {
  storage: Storage<S>;
  /**
   * @deprecated please migrate to {@link platformService}
   */
  notificationService: PlatformService<P>;
  keystoreService: KeystoreService;
  configService: ConfigService;
  internalService: InternalService;
  platformService: PlatformService<P>;
  backendProvider: BackendProvider;
  fullOwnershipService: OwnershipService;
  eventHub: EventHub;
}

``` 
Should I add a grantServiceField on it? is it will take security risk? @homura

homura: I forgot that we have not yet implemented `GrantService`, it looks like we can just read directly from the [config service](https://github.com/ckb-js/nexus/blob/22d6c8541e367cd3ad39a4d3c7f35b137ce44b24/packages/extension-chrome/src/rpc/walletImpl.ts#L21-L23)

homura: To get the current site, maybe we can add a `getActivedSiteInfo` method to `PlatformService`

```ts
declare function getActivedSiteInfo(): { host: string, favicon?: string } | undefined;
```

IronLu233: Yes I am doing this. your method name is more suitable.

packages/extension-chrome/src/pages/Popup/containers/Home.tsx
```diff
@@ -16,9 +18,37 @@ const FeedbackButton: FC<ButtonProps> = (props) => {
...
+  });
+
+  const whitelistSitesSet = useMemo(
+    () => new Set(whitelistQuery.data?.map((item) => item.host) || []),
+    [whitelistQuery.data],
+  );
+
+  const hasGrantedQuery = useQuery({
+    // Set in query key is not serialized correct, so add the whitelistSitesSet as a extra dependency
+    queryKey: ['platformService', whitelistSitesSet, whitelistQuery.data] as const,
```
homura, line 38: the `queryKey` in `ReactQuery` is a bit different from the `DepsList` in React, it is [hashed deterministically](https://tanstack.com/query/v4/docs/react/guides/query-keys#query-keys-are-hashed-deterministically). For `queryKey`, it is better to feed it the parameters used for query, e.g. `{id: X}`

For this case, we can just use a "getActiveSiteInfo" key is fine

```ts
useQuery(["getActiveSiteInfo"], async () => {
  await confitService.getWhitelist()

  ...
})
```

homura: If we think about it, to trigger a change to `whitelistQuery.data` we must first trigger a change to `queryKey: ['whitelist']` which means that `queryKey: ['whitelist']` is triggered in the same timing as `getActiveSiteInfo`

homura: to trigger a whitelist change, either invalidate the cache (cache timeout or manually clear it) or the query is actively triggered again (re-entering the page)

We currently do not manually invalidate the cache (perhaps not necessary), so the re-entry trigger alone is enough, i.e. the "getActiveSiteInfo" key is enough

homura: Additionally, if we want to manage the `queryKey` better, we can use the [factory](https://tanstack.com/query/v4/docs/react/community/lukemorales-query-key-factory) mode

IronLu233: Updated it. I merge two queries into one.

homura (approved):

PR #168, opened 2023-03-28
feat: supported get blockchain info rpc

### LINKED ISSUE #155: Find a reliable solution to enable dApps to obtain network
Nexus supports switching networks, but there is no particular way to get a network in a dApp.

We discussed using the `chain` in `get_blockchain_info` as a way to get the network, but we need to check if this is a reliable way to do it, or if there is a more reliable way

homura:
# What Changed

Added a method `ckb_getBlockChainInfo` to allow dApps determine which network Nexus is connected now

Resolve #155

## Motivation

Checkout the [comment](https://github.com/ckb-js/nexus/issues/155#issuecomment-1486924040)

## Change Type

- [ ] `documentation`
- [ ] `patch`
- [x] `minor`
- [ ] `major`

[nexus--canary.168.4544503195.zip](https://github.com/ckb-js/nexus/releases/download/v0.0.0-canary/nexus--canary.168.4544503195.zip)

### DIFF
--- a/packages/extension-chrome/src/services/ownership/backend/index.ts
+++ b/packages/extension-chrome/src/services/ownership/backend/index.ts
@@ -8,7 +8,9 @@ import chunk from 'lodash/chunk';
 import isEqual from 'lodash/isEqual';
 import { RPC as RpcType } from '@ckb-lumos/rpc/lib/types/rpc';
 import { NexusCommonErrors } from '../../../errors';
-import { createRpcClient, loadSecp256k1ScriptDep, toQueryParam, toCell, toScript } from './backendUtils';
+import { createRpcClient, loadSecp256k1ScriptDep, toCell, toQueryParam, toScript } from './backendUtils';
+import { ChainInfo } from '@ckb-lumos/base';
+import { ResultFormatter } from '@ckb-lumos/rpc';
 
 type GetLiveCellsResult = Paginate<Cell> & { lastLock?: Script };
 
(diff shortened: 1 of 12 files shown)


packages/extension-chrome/src/services/ownership/backend/index.ts
```diff
@@ -161,6 +165,11 @@ export function createBackend(_payload: { nodeUrl: string }): Backend {
       };
       return createTransactionSkeleton(tx, fetcher);
     },
+
+    getBlockchainInfo: async () => {
```
IronLu233, line 169: If the user have a pending network to add, should I validate the network? Is it exists an API for validating that the network is a CKB block chain RPC? Although this network config have not created yet.
image

IronLu233, line 169: If the user want to create a new network, and the network can not provide CKB RPC service, how can I validate it?

Because the new network has not been created yet. seems `getBlockchainInfo` not works for this case.

image

zhangyouxin: Looks like that a network should provide CKB RPC service. CKB nodes and light-clients(will) support that feature

I think Nexus can validate the user input `URL` before adding network to the list when the `Add` button is clicked

zhangyouxin (approved):

IronLu233 (approved):

PR #174, opened 2023-03-29
fix(extension-chrome): enable for http and local site 

### LINKED ISSUE #172: [enable_wallet] failed
nexus version : https://github.com/ckb-js/nexus/commit/f9f18a74bb3c60baddc3fab04590eaddd1498990
url: http://192.168.31.66:3000/
image

url: http://info.cern.ch/
image

homura:
## What Changes

Resolve #172 

## Change Type

- [ ] `documentation`
- [x] `patch`
- [ ] `minor`
- [ ] `major`

[nexus--canary.174.4555510584.zip](https://github.com/ckb-js/nexus/releases/download/v0.0.0-canary/nexus--canary.174.4555510584.zip)

### DIFF
--- a/packages/extension-chrome/__tests__/utils/validate.test.ts
+++ b/packages/extension-chrome/__tests__/utils/validate.test.ts
@@ -1,0 +1,30 @@
+import { validateHost } from '../../src/utils/validate';
+
+describe('validateHost', () => {
+  it('should be valid', () => {
+    expect(() => validateHost('localhost:1222')).not.toThrowError();
+    expect(() => validateHost('localhost')).not.toThrowError();
+    expect(() => validateHost('192.168.1.1')).not.toThrowError();
+    expect(() => validateHost('127.0.0.1')).not.toThrowError();
+    expect(() => validateHost('192.168.1.1:1233')).not.toThrowError();
+    expect(() => validateHost('nexus.com')).not.toThrowError();
+    expect(() => validateHost('nexus.com:1234')).not.toThrowError();
+    expect(() => validateHost('test.nexus.com:1234')).not.toThrowError();
+  });
+
+  it('should be invalid', () => {
+    expect(() => validateHost('')).toThrowError();
+    expect(() => validateHost(' ')).toThrowError();
+    expect(() => validateHost('localhost:')).toThrowError();
+    expect(() => validateHost('localhost: ')).toThrowError();
+    expect(() => validateHost('localhost: 1234')).toThrowError();
+    expect(() => validateHost('localhost:1234 ')).toThrowError();
+    // IP could not larger than 255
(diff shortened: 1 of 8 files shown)


packages/extension-chrome/src/contentScript/inpage.ts
```diff
@@ -1,14 +1,15 @@
...
 import { EventEmitter } from 'eventemitter3';
 import { isEventObject } from '../messaging/internal';
 import { createInjectedCkb } from '../injectedCkb';
-import { RpcClient, RpcMethods } from '@nexus-wallet/types';
+import type { RpcClient, RpcMethods } from '@nexus-wallet/types';
 
 const client = new JSONRPCClient(async (req) => {
   const response = await windowMessenger.sendMessage('contentAndInjected', req, 'content-script');
-  asserts.asserts(isJSONRPCResponse(response), `Invalid JSON-RPC response: ${response}`);
+  if (!isJSONRPCResponse(response)) {
```
zhangyouxin, line 10: Looks like an unnecessary change

packages/extension-chrome/__tests__/utils/validate.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('should be invalid', () => {
+    expect(() => validateHost('')).toThrowError();
+    expect(() => validateHost(' ')).toThrowError();
+    expect(() => validateHost('localhost:')).toThrowError();
+    expect(() => validateHost('localhost: ')).toThrowError();
+    expect(() => validateHost('localhost: 1234')).toThrowError();
+    expect(() => validateHost('localhost:1234 ')).toThrowError();
+    // IP could not larger than 255
```
zhangyouxin, line 22: maybe we should only allow [private network addresses](https://en.wikipedia.org/wiki/Private_network) when an IP address is detected:

- 192.168.0.0/16 
- 172.16.0.0/12 
- 10.0.0.0/8

Because the motivation of allowing ip addresses is only for DApp development when testing locally, real DApp production should not use a domain rather than IP address.

gpBlockchain (approved):

packages/extension-chrome/__tests__/utils/validate.test.ts
homura: If we disable the use of IP access, it will make the development inconvenient. Imagine we are working together, the contract engineer has developed the contract, the front-end engineer has developed the relevant pages, when they want to debug, they will generally use the private network for debugging, which is often the fastest way to use the private network IP address to debug

If a fishing dApp publishes a site using IP without using domain name, it may be a bit choking if we forbid IP access for such a rare case, at this stage we may consider how to make it convenient for developers, at a later stage, we may be able to make these functions into advanced configuration

zhangyouxin (approved):

Issue #191, opened 2023-04-10
Publish the public modules to npm


homura:
Nexus repo is a monorepo contains modules that need to be imported by other dApps, such as `protocol`, and `provider`. We need to publish these modules to npm for use as examples to run in CodeSandbox or for dApp to call

- intro a monorepo management tool, e.g. lerna
- canary release when PR
- release when merge

PR #194, opened 2023-04-10
refactor: make utils flatten

homura:
# What Changed

- flatted `utils` package
- correct `numeral` dep
- update deps

## Motivation

Before [publishing](https://github.com/ckb-js/nexus/issues/191) `utils` to npm registry, we should make the exported function be stable, so I have 

- removed unnecessary deps(`lumos/codec`) from utils to keep dependencies minimal
- flatted exported function to keep API clear

## Change Type

- [ ] `documentation`
- [ ] `patch`
- [ ] `minor`
- [ ] `major`


packages/types/src/base.ts
```diff
@@ -1,5 +1,7 @@
 export type Promisable<T> = PromiseLike<T> | T;
 
+export type Resolvable<T> = T | PromiseLike<T> | (() => T) | (() => PromiseLike<T>);
```
zhangyouxin, line 3:
```suggestion
export type Resolvable<T> = Promisable<T> | (() => Promisable<T>);
```

packages/utils/__tests__/asserts.ts
```diff
@@ -1,12 +1,6 @@
-import { asserts } from '../src';
+import { assert } from '../src';
 
 it('utils#asserts', () => {
-  expect(() => asserts.nonFalsy(0)).toThrow();
-  expect(() => asserts.nonFalsy('')).toThrow();
-  expect(() => asserts.nonFalsy(undefined)).toThrow();
-  expect(() => asserts.nonFalsy(null)).toThrow();
-
-  expect(() => asserts.nonFalsy(1)).not.toThrow();
```
zhangyouxin, line 9: I would replace with corresponding test cases

```
  expect(() => assert(0)).toThrow();
  expect(() => assert(1)).not.toThrow();
```

homura: supplied

packages/types/src/base.ts
homura: refactored

packages/utils/src/asserts.ts
```diff
@@ -1,15 +1,26 @@
...
+/**
+ * Assert condition is truthy, otherwise throw error.
+ * This function usually used with TypeScript's `asserts` type guard
+ * @example
+ *   assert(typeof value === 'string', 'value must be a string')
+ *   assert(typeof value === 'string', 'value must be a string, got %s', typeof value)
+ * @param condition
+ * @param args
+ */
+export function assert(condition: unknown, ...args: unknown[]): asserts condition {
```
IronLu233, line 13: Why not consider the implementation from community? It is more well-tested.

https://www.npmjs.com/package/assert

packages/utils/src/internal/format.ts
```diff
@@ -1,9 +1,10 @@
-import type { Provider } from '@nexus-wallet/types/lib/services/common';
-import type { Promisable } from '@nexus-wallet/types';
-import { bytes, BytesLike } from '@ckb-lumos/codec';
```
IronLu233, line 3: Seems implements bytes.hexify twice... Is it violate "Do not repeat your self" principle?

packages/utils/src/asserts.ts
homura: 1. keep the error message style consistent, e.g. "Assertion failed: 1 + 1 should be 2, but got 3  (version=0.0.1)"
2. it is a simple function, there is no need to introduce a library for a function

packages/utils/src/internal/format.ts
homura: only hexify for `Uint8Array` here, it is no need to import the entire `codec` library

zhangyouxin (approved):

IronLu233 (approved):

Issue #188, opened 2023-04-07
Cannot connect to nexus after opening the page for a long time


Keith-CY:
https://user-images.githubusercontent.com/7271329/230536403-30193689-7d7d-4fa1-aec6-ff4e8ccd4f7d.mov

After restarting nexus, the wallet can be connected, but the popup is still unavailable

Nexus build: https://github.com/ckb-js/nexus/releases/tag/v0.0.9

Keith-CY:
No error will be thrown on clicking <kbd>connect</kbd> when nexus is disabled
![image](https://user-images.githubusercontent.com/7271329/230537226-3a40b219-7dec-4ed2-8834-c3217e3f4a90.png)

homura:
Is it still in stuck? Could you help to dump the logs from `chrome://extensions`

<img width="769" alt="image" src="https://user-images.githubusercontent.com/7511174/230546166-6939fd96-60e7-4d8e-9180-f0f06b9a7a10.png">

Keith-CY:
The existing logs were wiped since nexus was disabled, but it is reproduced occasionally.



https://user-images.githubusercontent.com/7271329/230552457-4377b60f-86cb-46f8-b4ea-88c9142a1ea0.mov


Besides, I encountered another issue, the popup is blank after clicking <kbd>transfer</kbd>


https://user-images.githubusercontent.com/7271329/230552527-0a10b6e0-3b3d-43a5-8f6f-8f9c8d1c1fa7.mov

homura:
It is weird, I can't reproduce it in my environment

May I ask you how you install Nexus? Is installed by replacing the previous files? Or by dropping the unzipped folder directly under `chrome://extensions`?

If the answer is the first, there may indeed cause such a problem, you can try to reload Nexus after upgrade it manually

<img width="432" alt="image" src="https://user-images.githubusercontent.com/7511174/230556270-5e5f52b1-8f1f-4d2c-8a40-a4d28354c353.png">


But if the answer is the second, maybe if we just had a meeting, it would help us a lot

Keith-CY:
> It is weird, I can't reproduce it in my environment
> 
> May I ask you how you install Nexus? Is installed by replacing the previous files? Or by dropping the unzipped folder directly under `chrome://extensions`?
> 
> If the answer is the first, there may indeed cause such a problem, you can try to reload Nexus after upgrade it manually
> 
[...]

PR #209, opened 2023-04-13
chore: update webext-bridge to v6

zhangyouxin:
# What Changed

The [webext-bridge](https://github.com/zikaari/webext-bridge) package we use for messaging has been updated to v6, and fixed some bugs, this PR updates this dependency.

This dependency may also have some connection with #188 

## Change Type

- [ ] `documentation`
- [x] `patch`
- [ ] `minor`
- [ ] `major`


packages/extension-chrome/src/contentScript/content.ts
```diff
@@ -1,5 +1,5 @@
 import { isJSONRPCRequest, isJSONRPCResponse, JSONRPCClient } from 'json-rpc-2.0';
-import * as bridgeMessenger from 'webext-bridge';
+import { onMessage, sendMessage } from 'webext-bridge/content-script';
```
homura, line 2:
```suggestion
import * as bridgeMessenger from 'webext-bridge/content-script';
```

there are two messengers in this file, and to avoid confusion, it is best to distinguish between them with different names

zhangyouxin: updated

packages/extension-chrome/src/background/main.ts
```diff
@@ -3,7 +3,10 @@ import '../rpc/walletImpl';
 import '../rpc/debugImpl';
 import '../rpc/ckbImpl';
 import { createLogger, LIB_VERSION } from '@nexus-wallet/utils';
-import { Endpoint, onMessage } from 'webext-bridge';
+// eslint-disable-next-line import/no-unresolved
```
homura, line 6: why we should eslint disable here?

zhangyouxin: I think it's because we added some eslint rules recently, for `import/no-unresolved`, eslint may think `webext-bridge` comes from https://github.com/ckb-js/nexus/blob/main/packages/extension-chrome/src/types/messaging.d.ts#L5-L10 causing false positive report on importing `webext-bridge` everywhere

homura: can you try to help solve it?

zhangyouxin: turned off the rule

homura (approved):

PR #269, opened 2023-04-26
chore: fix incorrect coverage in code cov

homura:
# What Changed

- Run E2E testing before release
- Change the coverage reporter

## Motivation

The clover report is a bit strange, there are always some partial covered that are not well understood, I would like to try other reports to avoid this problem

image

### 

The PR reveals a more honest test coverage, so test coverage is reduced since the unit tests only covered the business code, the files like `PlatformService` can't be covered since it must run in a Chrome extension environment. 

According to this PR, I think we should 

- make the code more testable
  - separate business logic that is not related to the `browser` API for testing, e.g. `NotificationManager`, 
- refactor for high cohesion
  - putting together chrome-related adapters, e.g. [createBrowserExtensionEventHub](https://github.com/ckb-js/nexus/blob/c83c402371c874ffffa6d50e5ff5515623663ea2/packages/extension-chrome/src/services/event/extension.ts#L4), [makeBrowserExtensionModuleFactory](https://github.com/ckb-js/nexus/blob/c83c402371c874ffffa6d50e5ff5515623663ea2/packages/extension-chrome/src/services/factory/extension.ts#L9), [createBrowserExtensionStorage](https://github.com/ckb-js/nexus/blob/c83c402371c874ffffa6d50e5ff5515623663ea2/packages/extension-chrome/src/services/storage/extension.ts#L4), and ignore them in unit tests

## Change Type

- [ ] `documentation`
- [ ] `patch`
- [ ] `minor`
- [ ] `major`


homura:
I found that this _97.53%_ coverage is not real, we may need to define the `collectCoverageFrom` instead of statistics only for files with tests

IronLu233 (approved):

